PR #1, opened 2024-04-11
feat: added CI & CD, and docs deploy actions

jimezesinachi:
Why: we need a pipeline that tests every PR opened, and deploys the code when merged into the master branch.

What changed: updated the existing ci.yml, and added a new cd.yml (that publishes to NPM) & deploy.yml (that deploys the `docs` to GitHub Pages.

Important: ran Prettier through the repo (with the exception of the `static` folder in the `docs`, so you'll notice a bunch of formatting changes.


.github/workflows/cd.yml
```diff
@@ -0,0 +1,42 @@
+name: Continuous Deployment
+
+on:
```
fhur, line 3: Do we really need a separate cd.yml an deploy.yml? Can we not just have ci.yml and also run `yarn pub` there.

packages/.npmrc
```diff
@@ -1 +1,3 @@
 //registry.npmjs.org/:_authToken=${NPM_TOKEN}
+@synthql:registry=https://registry.npmjs.org
```
fhur, line 2: Why did you do these changes? Why are they neeed?

scripts/publish.cjs
```diff
@@ -10,49 +10,51 @@ function bumpMinorVersion(version) {
 const currentVersion = packageJson.version;
```
fhur, line 10: Maybe I wasn't clear but the idea is that we remove this file and move publishing into the nx.json config.

.github/workflows/cd.yml
jimezesinachi: Not exactly. If you look, I have modified the `ci.yml` to trigger ONLY on PRs opened against `master`, no longer also on pushes to `master`. This is because we do not want to trigger a publish to npm ANYTIME someone opens a PR against master. So we have to create a seperate workflow for when changes are successfully merged into `master`.

The `cd.yml` is what runs now on pushes and/or merges into `master`. Since we only want to deploy the full changes, as is merged in master, it makes sense to separate them this way. So, this way, `ci.yml` is focused on making sure that PRs are good enough to be merged, while `cd.yml` triggers a new release once any such changes are succcessfully merged into main. This also allows us to detect if the changes merged into main include new failing tests or broken code that we couldn't in the PR run.

.markdownlint.json
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
jimezesinachi: Understood! 👍🏾

packages/.npmrc
jimezesinachi: Got this from the GitHub documentation for publishing to NPM from actions. Unclear if the publish won't work without the two new lines, but I will remove them as needed if there's a failure.

scripts/publish.cjs
jimezesinachi: I understand that now. Will work on doing that now!

fhur:
Much cleaner diff, thanks @jimezesinachi !

.github/workflows/cd.yml
fhur: OK makes sense 👍

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
+name: Deploy documentation to GitHub Pages
+
+on:
+    push:
+        branches:
+            - main
```
fhur, line 6: there's no main branch 🤔 typo?

jimezesinachi: Correct! Wiil fix that now!

PR #3, opened 2024-04-16
feat: added @synthql/cli package

jimezesinachi:
# Why

This CLI will allow users access the generation and introspection features of the core library

# What changed

New feature

# Important

There's currently an issue with the `tsconfig.json` and/or build/module configuration. Basically, the source code builds, but the installed packages don't appear in the `node_modules` (not there, and not symlinked from the root), and hence the script fails (with a `Error: Cannot find module '@synthql/backend'`). It might be some quirk with Yarn, but I don't know for sure. Need help debuggging it.


packages/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+            connectionString: {
+                alias: 'url',
+                default: 'postgres://postgres:postgres@localhost:5432/postgres',
+            },
+            outputDir: {
+                alias: 'out',
+                default: 'src',
+            },
+            defaultSchema: {
+                alias: 'default',
```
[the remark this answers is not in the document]
fhur: In fact here's my rule of thumb for command/param aliases:

1. If the command/param is infrequently invoked (i.e. typed) by the user, prefer longer more descriptive names. 
2. If the command is often typed manually by the user, add one or two letter aliases. Example: `git status` => `git s`
3. Never add synonyms or more than one name to a command or param. If people find examples online this will eventually just lead to confusion.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+                schemas: argv.schemas,
+            });
+
+            console.info(result);
+        },
+    )
+    .array('schemas')
+    .help()
+    .demandCommand(1)
+    .parse();
```
fhur, line 42: This is also missing unit/e2e tests

fhur (commented):
Lots of good progress @jimezesinachi, good job!

Made a few comments :)

packages/cli/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@synthql/cli",
+    "version": "0.0.1",
```
fhur, line 3: use the same version number as the other packages

fhur:
With regards to the structure of the test files, please follow existing conventions: Concretely, keep the test files in the same folder as the file being tested.

Example:

```
src/commands/foo.ts
src/commands/foo.test.ts
```

Don't add a `test` folder

yarn.lock
```diff
@@ -12273,6 +12273,11 @@ typescript@^5:
   resolved "https://registry.npmjs.org/typescript/-/typescript-5.3.3.tgz"
   integrity sha512-pXWcraxM0uxAS+tN0AG/BF2TyqmHO014Z070UsJ+pFvYuRSq8KH8DmWpnbXe0pEPDHXZV3FcAbJkijJ5oNEnWw==
 
+typescript@^5.4.5:
```
fhur, line 12276: Note: You're adding another typescript version. You should be using the same one used by the other modules.

fhur: In fact this is probably a good moment to pick the same version across the docs and ui module. I would go for `^5.4.5`

packages/cli/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "repository": {
+        "type": "git",
+        "url": "git@github.com:synthql/synthql.git"
+    },
+    "license": "MIT",
+    "scripts": {
+        "build": "yarn tsc",
+        "publish:minor": "yarn publish --access public --no-git-tag-version",
+        "format": "yarn prettier --config ../../prettier.config.js --write ./src/"
```
fhur, line 12: This is missing a `test:ci` and `test` script

packages/cli/package.json
fhur: We should try to make sure all packages in the monorepo all have the same version. This makes it easier for users to figure out which package versions are compatible with each other: "just make sure they are all using the same version"

packages/cli/package.json
```diff
@@ -0,0 +1,28 @@
...
+        "yargs": "^17.7.2"
+    },
+    "devDependencies": {
+        "@types/node": "^20.10.5",
+        "@types/pg": "^8.10.9",
+        "@types/yargs": "^17.0.32",
+        "typescript": "^5.4.5"
+    },
+    "bin": {
+        "synthql": "build/index.js"
```
fhur, line 26: You're not actually writing anything to build/index.js

packages/cli/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@synthql/cli",
+    "version": "0.0.1",
+    "repository": {
+        "type": "git",
+        "url": "git@github.com:synthql/synthql.git"
+    },
+    "license": "MIT",
+    "scripts": {
+        "build": "yarn tsc",
```
fhur, line 10: Please be consistent with other packages unless there is a good reason to deviate: We use `tsc` just for type checking and `vite` for building/bundling.

If you look at the package.json of other packages you will notice

jimezesinachi: Indeed. I did do this at the beginning, but I was running into some errors due to Vite struggling with bundle sharing/compatibility issues, showing some errors about how some of the modules weren't available outside of the browser. But I will set it up again now, let's loot at fixing the errors.

packages/cli/package.json
jimezesinachi: Correct. I just relaized that the root .`gitignore` was excuding all `build/` folders on commits. I will fix that now!

packages/cli/package.json
jimezesinachi: Adding this now using the Vite commands from the `queries` package.

packages/cli/dist/commands/generate/index.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
fhur, line 1: @jimezesinachi please be more careful with your MR: this file shouldn't have been commited

fhur: The same is true for a bunch of other .js/.map.js files below

packages/cli/src/index.ts
fhur: You haven't implemented tests yet

jimezesinachi: Yes. I am looking to fix the build issue first before proceeding

packages/cli/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+                        'Connection URL to the database to generate a schema from',
+                    alias: 'url',
+                    default:
+                        'postgres://postgres:postgres@localhost:5432/postgres',
+                })
+                .option('out', {
+                    type: 'string',
+                    description:
+                        'Output directory of the generated schema file',
+                    default: 'src',
```
fhur, line 25: Let's change this default to src/generated/synthql

packages/cli/src/commands/generate/index.ts
```diff
@@ -0,0 +1,33 @@
...
+        connectionString,
+    });
+
+    const queryEngine = new QueryEngine({
+        pool,
+        schema: defaultSchema,
+    });
+
+    const result = await queryEngine.generateSchema({
+        out,
```
fhur, line 27: Given that `out` is a folder, you should append a file name here, such as `schema.ts`

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,45 @@
...
+                })
+                .option('defaultSchema', {
+                    type: 'string',
+                    description:
+                        'Default schema to be included in the generation',
+                })
+                .option('schemas', {
+                    description:
+                        'List of schemas to be included in the generation',
+                    default: [] as string[],
```
fhur, line 33: Isn't this a bad default? Doesn't this mean that by default you include no schemas?

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,45 @@
...
+                    description:
+                        'Connection URL to the database to generate a schema from',
+                    alias: 'url',
+                    default:
+                        'postgres://postgres:postgres@localhost:5432/postgres',
+                })
+                .option('out', {
+                    type: 'string',
+                    description:
+                        'Output directory of the generated schema file',
```
fhur, line 22:
```suggestion
                        'The path to the folder where the generated files will be created.',
```

packages/cli/src/cli.ts
jimezesinachi: Not exactly. The `defaultSchema` is always set, and its value is `public`. So the idea is, the `schemas` array should be set ONLY when you want to include more than one schema. Otherwise you should set the `defaultSchema` option, or use the default value for `defaultSchema` (i.e. `public`).

.markdownlint.json
```diff
@@ -5,8 +5,8 @@
         "line_length": 250
     },
     "list-marker-space": {
-        "ol_single": 3,
-        "ol_multi": 3,
+        "ol_single": 2,
```
fhur, line 8: Once again: please don't mix changes. For now it's OK to leave these changes in, but try not to make this again in the future.

This makes the diff unnecessarily large, and it's also a sign of loss of focus. If you're working on the CLI there's no good reason to tweak these values.

packages/cli/src/cli.test.ts
```diff
@@ -0,0 +1,275 @@
...
+
+        beforeEach(() => {
+            vitest.resetModules();
+            vitest.resetAllMocks();
+        });
+
+        it(
+            'Generate schema types with the CLI options default values',
+            async () => {
+                const outDir = `${out}/${Date.now()}-${randomUUID()}`;
```
fhur, line 39: While this is totally fine, you could have just done randomUUID or even just Math.random(). The probability of a colission is super low.

fhur: No need to change anything: just an informational message.

packages/cli/src/cli.test.ts
```diff
@@ -0,0 +1,275 @@
...
+
+                const argv = ['node', 'generate-default.js', 'generate'];
+
+                // Compose file path
+                const outFilePath = path.join(outDir, `db.ts`);
+
+                // Check that file with the name doesn't exist yet
+                assert(!fs.existsSync(outFilePath));
```
fhur, line 47: Instead of using the assert module you should rely on normal expect() assertions.

packages/cli/src/generated/synthql/db.ts
```diff
@@ -0,0 +1,60 @@
+// This file was generated by @synthql/backend
+import { query } from '@synthql/queries';
```
fhur, line 2: @jimezesinachi this shouldn't have been commited

fhur (commented):
Almost there, good job @jimezesinachi !

packages/cli/src/generated/synthql/db.ts
fhur: Pay more attention to your diffs. This is a sign that you're not reviewing your own code.

.markdownlint.json
fhur: I also noticed that you made a bunch of changes to the docs, which I actually really appreciate :)

But again: focus focus focus. It's all about focus! We'll work on improving the docs later.

jimezesinachi: Makes sense, but in this case, the changes were necessary for updating the documentation.

packages/cli/src/generated/synthql/db.ts
jimezesinachi: I was thinking the same, but I did when I noticed the generated files had been committed in the backend package.

packages/cli/src/cli.test.ts
fhur: yeah

jimezesinachi: Updated!

packages/cli/src/generated/synthql/db.ts
fhur: gotcha. In the case of other files we do rely on them for testing, but not here.

jimezesinachi: Deleted!

packages/cli/src/cli.test.ts
```diff
@@ -44,13 +43,13 @@ describe('CLI tests', () => {
                 const outFilePath = path.join(outDir, `db.ts`);
 
                 // Check that file with the name doesn't exist yet
-                assert(!fs.existsSync(outFilePath));
+                expect(!fs.existsSync(outFilePath));
```
fhur, line 46: @jimezesinachi pay more attention: you're not `expect`-ing enything 👀 

Also below

PR #4, opened 2024-04-26
fix: added `sql-formatter` as normal dependency to `backend` package

jimezesinachi:

## Why

To fix the issue with the CLI breaking due to this package missing from the backend

## What changed

Added a single dependency

## Important

Nothing much.


fhur (approved):
LGT

PR #7, opened 2024-05-02
feat: added `Express` request handler

jimezesinachi:

## Why

Adding an Express request handler to allow easier setup and usage in Express applications

## What changed

A single handler function and unit tests

## Important

Tests are still very rudimentary at this point. All commented code is failing tests that I'm working on, and is there intentionally, until final review is requested.


packages/backend/src/tests/pagilaDb.ts
```diff
@@ -0,0 +1,1596 @@
+// This file was generated by @synthql/backend
+import { query } from '@synthql/queries';
```
fhur, line 2: @jimezesinachi there is already a pagila DB schema in the /backend package, no need to create a new one.

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
```diff
@@ -0,0 +1,58 @@
+import { beforeAll, describe, expect, test } from 'vitest';
+import { createExpressSynthqlHandler } from './createExpressSynthqlHandler';
+import { QueryEngine } from '../..';
+import { getMockReq, getMockRes } from 'vitest-mock-express';
```
fhur, line 4: Do you really need an additional library for this?

fhur: General note: Your default position shouldn't be to add libs if you're in need of something. Dependencies are a dangerous thing. They often bring hidden complexities that you often don't realise until it's too late to remove them.

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { DB, from } from '../../tests/pagilaDb';
+
+describe('createExpressSynthqlHandler', () => {
+    let queryEngine: QueryEngine<DB>;
+
+    beforeAll(async () => {
+        queryEngine = new QueryEngine<DB>({
```
fhur, line 12: There's a global QueryEngine instance that you can reuse

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+                returnLastOnly: headers['x-return-last-only'] === 'true',
+            })) {
+                res.write(JSON.stringify(intermediateResult) + '\n');
+            }
+
+            res.end('SynthQL request handled successfully!');
+        } catch (error) {
+            res.write(JSON.stringify({ error: String(error) }) + '\n');
+
+            res.end('SynthQL request failed with error!');
```
fhur, line 28: What's the reasoning for this?

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+        const query = JSON.parse(req.body);
+
+        res.status(200);
+        res.setHeader('Content-Type', 'application/x-ndjson');
+
+        try {
+            for await (const intermediateResult of queryEngine.execute(query, {
+                returnLastOnly: headers['x-return-last-only'] === 'true',
+            })) {
+                res.write(JSON.stringify(intermediateResult) + '\n');
```
fhur, line 21: Concatenating ad this level is creating an unnecesay copy of the entire JSON string.

Write the lines separately e.g. res.write(json); res.write('\n')

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
fhur: If I'm not being mistaken this is incompatible with the JSONL format on the react client.

packages/backend/src/tests/pagilaDb.ts
jimezesinachi: That's correct! But I noticed that the schema generated was old, so I figured that I'll generate the new one for the new tests, then update the old tests to use the new one over time until we can delete the old schema.

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export async function createExpressSynthqlHandler<T>(
+    queryEngine: QueryEngine<T>,
+): Promise<ExpressSynthqlHandler> {
+    return async (req, res) => {
+        const headers = req.headers;
+
+        const query = JSON.parse(req.body);
+
+        res.status(200);
```
fhur, line 14: When setting `x-return-last-only` you shouldn't return 200 always as you will know in some situations when an error happens.

fhur:
> Tests are still very rudimentary at this point. All commented code is failing tests that I'm working on, and is there intentionally, until final review is requested.

I don't understand the reasoning for this @jimezesinachi. How else do you know if it's working properly?

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
jimezesinachi: Understood. I did try to create it without, but it meant that we would have to remove the types on `req` and `res` in the handler function, or use partial types, both of which didn't seem ideal.

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
jimezesinachi: Understood.  Can you point me to it?

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
jimezesinachi: The idea was to listen for `res.end` to be called with that value, but you're right that it's not consistent with the `nl-json` format. Will remove that now.

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
jimezesinachi: Understood. On it!

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
jimezesinachi: Makes sense.

jimezesinachi:
> > Tests are still very rudimentary at this point. All commented code is failing tests that I'm working on, and is there intentionally, until final review is requested.
> 
> I don't understand the reasoning for this @jimezesinachi. How else do you know if it's working properly? 

What I'm saying is, all the current tests work and pass, and any commented code is work in progress (and hence you don't need to point them out, similar to what I mentioned last week).

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
fhur: https://github.com/synthql/synthql/blob/874b522cbd60d3b15061c4ef6efa90e80d0acfc2/packages/backend/src/tests/queryEngine.ts

packages/backend/src/tests/pagilaDb.ts
fhur: Nah, better to do that in a separate change.

fhur: As a general rule: try to avoid intermediate dirty stages.

fhur: Every change you make to the repo should always leave the repo in a better stage. This isn't always possible, but it should be the default way of working.

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
fhur: General note: this is a sign that you're not peeking into the repo. Working on a new codebase can sometimes be daunting, but the only way to get to a point where you feel familiar with the codebase is by looking into it.

A strategy that I sometimes follow is to periodically look into the codebase from a high level and slowly descend into lower levels the more you get familiar with it. Here's some concrete ideas:

High level:
1. Start by looking at the file/folder structure. Notice patterns and naming.
1. Understand the package structure. Which packages depend on which.
1. Read the docs. Notice common re-ocurring terms.
1. Read/skim the [API docs](https://synthql.github.io/SynthQL/reference/)

At this stage you're just learning the current structure, patterns and learning terminology/naming used in the project.

You're also learning where certain things are in the codebase.

Mid level:
1. Read the test/coverage reports: What things are being tested, which files have high coverage, low coverage, etc.
1. Skim the test files. 
2. Skim the type definitions.

At this stage you should already be familiar with the naming, conventions, patterns and you should have identified the most important classes/functions (e.g. QueryClient, useSynthQL, etc.) You should also be able to explain to your grandma what they do and how they relate to each other.

Low level:
1. Understand the flow of information: which functions call which functions.
2. Understand the main algorithms: e.g. how is planning implemented.

At this stage you should be able to tell how information flows throughout the project (e.g. what are the most important stages, which classes/functions are public API vs internal, how/when do they communicate with each other)


Some tools/techniques to jumpstart this process:
1. Write tests. This will force you to understand the inputs/outputs.
4. Make small refactors. This will force you to read the codebase, and get a sense of completion. Keep the refactors small as they might break things that you don't understand yet.
5. Expand the documentation: If there are bits that you don't understand, write about them and then document your writings in the docs so that others will be able to benefit from this knowledge.

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
fhur: Notice that the client expects every line to be JSON so this would have broken the client. This would have been caught by a test.

packages/backend/src/tests/pagilaDb.ts
jimezesinachi: Understood. Will update now!

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
```diff
@@ -0,0 +1,49 @@
+import { describe, expect, test } from 'vitest';
+import { createExpressSynthqlHandler } from './createExpressSynthqlHandler';
+import { queryEngine } from '../../tests/queryEngine';
+import { getMockReq, getMockRes } from 'vitest-mock-express';
+
+import { from } from '../../tests/generated.schema';
+import { DB } from '../../tests/db';
+
+describe('createExpressSynthqlHandler', () => {
```
fhur, line 9: This test isn't really testing anything useful @jimezesinachi 

A few of the important invariants you should be testing:

1. The response is always 1..n lines of JSON
2. when returnLastOnly the status code can vary depending on the response kind
3. There's really no need to mock: spin up an actual server using `http.createServer`
4. Tests for error cases i.e. when the query is malformed
5. Add also tests using the react client. Notice that there is already a http.createServer in /react, instead of the custom handler code refactor to use the createExpressSynthqlHandler

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,33 @@
+import { QueryEngine } from '../..';
+import { Request, Response } from 'express';
```
fhur, line 2: Use `import type`, we don't need any runtime dependency on express

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type ExpressSynthqlHandler = (req: Request, res: Response) => void;
+
+export async function createExpressSynthqlHandler<T>(
+    queryEngine: QueryEngine<T>,
+): Promise<ExpressSynthqlHandler> {
+    return async (req, res) => {
+        const headers = req.headers;
+
+        const query = JSON.parse(req.body);
```
fhur, line 12: This can fail: add proper error handling

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
fhur: There's a better way: see my comment about http.createServer below. So I think we can get rid of vitest-mock-express

packages/react/src/test/appendBody.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function appendBody(expressHandler: ExpressSynthqlHandler) {
+    return async (req: IncomingMessage, res: ServerResponse) => {
+        const body = await readBody(req);
+
+        const newReq = {
+            ...req,
+            headers: req.headers,
+            body: body,
+        } as IncomingMessageWithBody;
```
fhur, line 40: don't do `as Type`, this is generally unsafe behaviour. Instead do

```ts
 const newReq: Type = { ... }
```

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, test } from 'vitest';
+import { query } from '@synthql/queries';
+
+import { DB } from '../../tests/db';
+const from = query<DB>().from;
+
+describe('createExpressSynthqlHandler', () => {
+    test(`Query execution is successful`, async () => {
+        const q = from('actor')
```
fhur, line 9: this isn't testing anything @jimezesinachi

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+                    res.write(JSON.stringify(intermediateResult));
+
+                    res.write('\n');
+                }
+
+                res.end();
+            }
+        } catch (error) {
+            res.statusCode = 400;
+
```
fhur, line 68: Why

so

much

space

between

lines? 😅 .

Makes

it

harder

to read.

packages/react/src/test/appendBody.ts
jimezesinachi: This would be ideal, but the goal here is actually type casting. We don't have all the values for the type, so using a type annotation in this case, causes a type mismatch error.

packages/backend/src/util/handlers/createExpressSynthqlHandler.test.ts
jimezesinachi: Correct! These are just the placeholder logic for whatever tests we decide to add for the handler function.

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
jimezesinachi: Noted and fixed!

PR #8, opened 2024-05-09
feat: updating `queries` API types

jimezesinachi:

## Why

We have updated the way we generate the database types. This warrants an update to the the way we generate the types for the `queries` API. This PR begins this update.

## What changed

Changed the implementation for the type-safety generator types in the `queries` package

## Important


packages/queries/src/types/ColumnReference.ts
```diff
@@ -5,8 +5,16 @@
...
     [TTable in keyof DB]: DB[TTable] extends object
-        ? `${TTable & string}.${(keyof DB[TTable] & string) | '*'}`
+        ? `${TTable & string}.${
+              | ((DB[TTable] extends {
+                    columns: infer C;
+                }
+                    ? keyof C
+                    : never) &
+                    string)
+              | '*'}`
```
fhur, line 18: let's remove the '*', we don't really support that for now

fhur:
Good job @jimezesinachi, changes look good so far👍

packages/backend/src/execution/execute.test.ts
```diff
@@ -1,15 +1,33 @@
...
 
 interface DbWithVirtualTables extends DB {
-    'virtual.film_rating': {
-        film_id: number;
-        rating: string;
+    film_rating: {
+        columns: {
+            film_id: {
+                type: PgCatalogInt4;
+                // selectable: true;
```
fhur, line 15: these types are wrong: they should be isPrimaryKey: true, nullable:false, etc.

packages/backend/src/execution/executors/PgExecutor/PgExecutor.test.ts
```diff
@@ -1,26 +1,28 @@
...
 import { describe, expect, it } from 'vitest';
 import { PgExecutor } from '.';
-import { from } from '../../../tests/generated.schema';
+import { DB } from '../../../tests/generated';
 import { pool } from '../../../tests/queryEngine';
 import { createRefContext } from '../../../refs/RefContext';
 import { QueryProviderExecutor } from '../QueryProviderExecutor';
 import { collectColumnReferences } from '../../../query/collectColumnReferences';
 
+const from = query<DB>().from;
```
fhur, line 10: You should modify the generator logic to append this to the /introspection index generation logic so you don't have to keep re-typing this everywhere. Something like:

```ts
fs.writeFileSync('index.ts',`const from = query<DB>().from;`)


```

packages/backend/src/query/collectColumnReferences.test.ts
```diff
@@ -1,28 +1,30 @@
...
+
 describe('collectColumnReferences', () => {
     describe('collectColumnReferences', () => {
         const cases: Array<{
             query: AnyQuery;
             expected: Array<string>;
         }> = [
             {
-                query: from('public.actor').where({ actor_id: 1 }).many(),
+                query: from('actor').where({ actor_id: 1 }).many(),
```
fhur, line 19: For future PRs: would have been nice if you would have done this change first. Most of the diffs in this MR are just public.x to x.

Long diffs makes it harder to spot bugs.

packages/backend/src/tests/compareCustomers.ts
```diff
@@ -1,7 +1,10 @@
-import { DB } from './generated.schema';
+import { DB } from './generated';
+import { TableTypes } from './getTableTypes';
 
-type Customers = Pick<DB['public.customer'], 'customer_id'>;
+type C = DB['customer']['columns'];
```
fhur, line 4: what is `C`? can you come up with a better name?

packages/backend/src/tests/getTableTypes.ts
```diff
@@ -0,0 +1,9 @@
+export type TableTypes<Table> = {
```
fhur, line 1: add some documentation (and example) of what this type does.

packages/backend/src/tests/getTableTypes.ts
```diff
@@ -0,0 +1,9 @@
...
+    [TTable in keyof Table]: Table[TTable] extends object
```
fhur, line 2: I think this type can be simplified: do you really need extends object and then extends { type: infer C }?

Also `C` is a bad name. Prefer the convention T<DescriptiveName>

packages/backend/src/tests/getTableTypes.ts
fhur: Also not sure I like the naming as I think (but I'm not sure) that this returns columns, not tables 🤔

packages/queries/src/generated/types.test.ts
```diff
@@ -0,0 +1,47 @@
+import { query } from '../query';
+import { Column } from '../types/Column';
+import { ColumnValue } from '../types/ColumnValue';
+import { DB } from '.';
+import { describe, test } from 'vitest';
+import { col } from '../col';
+import { ColumnReference } from '../types/ColumnReference';
+
+describe('New types tests', () => {
```
fhur, line 9: I guess we can remove this file now, all of these cases are covered by existing tests I think

packages/queries/src/types/Column.ts
```diff
@@ -1,9 +1,13 @@
...
 /**
- * The name of a column in the database.
+ * The columns of a table.
  *
  * @param TTable The table the column belongs to.
  */
 
-export type Column<DB, TTable extends Table<DB>> = keyof DB[TTable] & string;
+export type Column<DB, TTable extends Table<DB>> = DB[TTable] extends {
+    columns: infer C;
```
fhur, line 10: Use a more descriptive name e.g. TColumnDef

packages/queries/src/types/Column.ts
```diff
@@ -1,9 +1,13 @@
 import { Table } from './Table';
 
 /**
- * The name of a column in the database.
+ * The columns of a table.
```
fhur, line 4: Please add an example

```
Column<B,'films'> = ...
```

packages/queries/src/types/ColumnValue.ts
```diff
@@ -14,15 +14,11 @@ export type ColumnValue<
...
-          ? T | null
-          : DB[TTable][TColumn];
+> = DB[TTable] extends { columns: infer TColumnDef }
+    ? ColumnType<TColumnDef, TColumn>
+    : never;
 
-type Selectable<T> = {
-    readonly __select__: T;
-};
+type ColumnType<
```
fhur, line 21: Please add some documentation

fhur:
This is really good progress, great job @jimezesinachi 👍  Definitely moving faster than I was expecting

packages/queries/src/types/ColumnValue.ts
```diff
@@ -14,15 +14,11 @@ export type ColumnValue<
...
+    TColumnDef,
+    TColumn extends keyof TColumnDef,
+> = TColumnDef[TColumn] extends { type: infer TType } ? TType : never;
```
fhur, line 24: There's likely a bug here: this isn't taking nullable columns into account. Should be something like

Case 1:
{type: infer TType, isNullable: true} ? null | TType

Case 2:
{type: infer TType, isNullable:false } ? TType

Case 3:
never

packages/backend/src/tests/getColumnDataTypes.ts
```diff
@@ -0,0 +1,24 @@
...
+ * const customer: CustomerColumnDataTypes = {
+ *   customer_id: 1;
+ *   store_id: 3;
+ *   first_name: 'First';
+ *   last_name: 'Last';
+ * }
+ * ```
+ */
+
+export type ColumnDataTypes<Table> = {
```
fhur, line 18: This is not taking nullability into consideration

packages/introspect/src/generate.ts
```diff
@@ -17,18 +17,24 @@ export async function generate({
     defaultSchema,
     outDir,
     formatter = async (str) => str,
+    config = {
```
fhur, line 20: don't wrap this in an additional `config` object. The wrapping is adding no additional value. The config object is already wrapping all the properties

packages/queries/src/types/Column.ts
```diff
@@ -1,13 +1,20 @@
 import { Table } from './Table';
 
 /**
- * The columns of a table.
+ * The names of the columns of a table, returned as a string intersection type.
```
fhur, line 4: this is not a string intersection, its a string union

packages/introspect/src/generate.ts
fhur: Also note that there is actually a bug here: you can pass `config = {}` which will not default to config = {importLocation}

fhur: Given that this should not be used by normal users I would suggest using a more explicit name like`dangerous_importQueriesLocation`

packages/queries/src/types/ColumnValue.ts
fhur: You changed the documentation for ColumnValue (which I thought was already good enough) but you didn't add any do ColumnType which was my request 😅

packages/queries/src/types/ColumnValue.ts
```diff
@@ -21,4 +27,13 @@ export type ColumnValue<
 type ColumnType<
     TColumnDef,
     TColumn extends keyof TColumnDef,
-> = TColumnDef[TColumn] extends { type: infer TType } ? TType : never;
+> = TColumnDef[TColumn] extends
+    | { type: infer TNullableType; nullable: true }
+    | undefined
```
fhur, line 32: why `| undefined`? can you explain this case?

packages/queries/src/types/Where.ts
```diff
@@ -7,11 +7,11 @@ import { WhereClause } from './WhereClause';
...
  * This is currently provides _very_ basic support for filtering.
  *
- * Currently, this can be either
+ * Currently, this can be either:
  *
- * 1. `{column: {operator: value}}` example: `{age: {'>': 18}}`, which translates to `age > 18`
- * 1. `{column: value}` which is equivalent to `{column: {'=': value}}`
- * 1. `{column: col('table.column')}` which translates to `column = table.column`
+ * 1. `{column: {operator: value}}` example: `{age: {'>': 18}}`, which translates to `age > 18`.
+ * 2. `{column: value}` which is equivalent to `{column: {'=': value}}`.
```
fhur, line 13: Note that the 1. 1. 1. is intentional: Markdown is smart and will render them as 1. 2. 3. 

It's just a convenience so in case you add a new column at the second line you don't have to update 2 -> and 3 -> 4.

Doesn't really matter for small lists but for a large numbered list it starts to make a difference.

No need to change anything btw. Just an informational note.

packages/queries/src/types/ColumnValue.ts
jimezesinachi: Ahh.. that explains it, I didn't realize you were referring to ColumnType. Will do so now!

packages/queries/src/types/ColumnValue.ts
fhur: Oh that looks like a bug in the types.

fhur: Can you fix it? Should be a really easy fix.

jimezesinachi: Ahh.. okay. On it

PR #9, opened 2024-05-17
feat: default select all columns

jimezesinachi:

## Why

Currently, when using the `query` API, there's no mechanism that autoselects all selectable fields in the database. This PR adds this feature.

## What changed

A change to the `query` API implementation and a minor change to the `QueryBuilder`.


packages/queries/src/query.ts
```diff
@@ -8,9 +8,9 @@ export class QueryBuilder<
     DB,
     TTable extends Table<DB>,
     TWhere extends Where<DB, TTable>,
-    TCardinality extends 'one' | 'maybe' | 'many',
-    TInclude extends Include<DB>,
     TSelect extends Select<DB, TTable>,
+    TInclude extends Include<DB>,
```
fhur, line 12: As I've mentioned before @jimezesinachi, try to avoid these refactorings. They just make the PR longer to review but don't really add anything.

packages/queries/src/query.ts
```diff
@@ -325,19 +325,71 @@ export class QueryBuilder<
     }
 }
 
-export function query<DB>() {
+export function query<DB>(schema: any) {
```
fhur, line 328: `schema: any` is not an option. Please define and use a concrete type.

packages/queries/src/query.ts
```diff
@@ -325,19 +325,71 @@ export class QueryBuilder<
...
     return {
         from<TTable extends Table<DB>>(table: TTable) {
+            type TKeys = Array<Column<DB, TTable>>;
```
fhur, line 331: This whole blob of code is way too dirty.

What you should do is create a function that given a schema returns all the columns 

e.g.

`getColumns(schema: Schema)`

This will make it easier to read, but more importantly to write unit tests for it.

fhur (commented):
Good progress @jimezesinachi 👍 

Added some comments about cleaning up the column extraction logic.

packages/queries/src/query.ts
jimezesinachi: I agree! But this one was really necessary. The difference in the argument order caused some confusion and time when I was working on the `from`. I only changed it because of that.

packages/queries/src/getColumns.ts
```diff
@@ -0,0 +1,33 @@
+import { DbSchema } from './types/schemas';
+
+export function getColumns(schema: DbSchema, table: string) {
```
fhur, line 3: Without any docs I would have assumed that a function called `getColumns(schema,table)` would return all the columns. That's not the case, though, it returns the selectable columns.

Naming your reflect the particularities of your function. In this case the important bit is that it's returning _only_ the columns that are selectable.

packages/queries/src/getColumns.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const tableDef = properties[table];
+
+    const tableDefProperties = tableDef.properties;
+
+    const columnsDef = tableDefProperties.columns;
+
+    const columnsDefProperties = columnsDef.properties;
+
+    for (const column in columnsDefProperties) {
```
fhur, line 16: This can be simplified using Object.values

 `for (const columnDef of Object.values(columnsDefProperties))`

This avoids the `if (column)` check, and avoids the `columnDef = columnsDefProperties[column]`

packages/queries/src/getColumns.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (column) {
+            const columnDef = columnsDefProperties[column];
+
+            const columnDefProperties = columnDef.properties;
+
+            const selectableDef = columnDefProperties.selectable;
```
fhur, line 22: There's no need to declare `selectableDef`. This is just adding an extra line of code with no purpose. 

const col

As a general rule of thumb don't declare additional variables unless there's a reason for it.

packages/queries/src/query.ts
```diff
@@ -3,6 +3,8 @@ import { Where } from './types/Where';
 import { Select } from './types/Select';
 import { Column } from './types/Column';
 import { Table } from './types/Table';
+import { DbSchema } from './types/schemas';
```
fhur, line 6: This is not following conventions @jimezesinachi.

Notice that all other types in /queries are single-type-per-file and the file is named identically as the exposed type.

packages/react/src/test/echoSchema.ts
```diff
@@ -0,0 +1,28 @@
+export const schema = {
```
fhur, line 1: Don't put this in another file. This file is _really_ coupled to the echoDb.ts file, hardly other file will ever use it, so better to keep it private there.

packages/queries/src/getColumns.ts
fhur: Two suggestions:

1. Add some docs (ideally with an example)
2. Rename the function to `getSelectableColumns`

fhur, line 3: With regards to the implementation: this can be cleaned up much more if you have the right building blocks:

Remember what I said about divide and conquer? Split the problem into 3-4 basic functions.

- getSelectableColumns(schema, table)
- getTableDef(schema, table)
- isSelectableColumn(columnDef):boolean

```ts
// pseudocode
getSelectableColumns(schema,table) {
    const table = getTableDef(schema,table);
    for (const column of getColumnDefs(table)) {
         if (isSelectableColumn(column)) {
              // TODO
         }
    }
}
```

packages/react/src/useSynthql.test.tsx
```diff
@@ -56,6 +56,32 @@ describe('useSynthql tests', () => {
         pagilaServer?.server.close();
     });
 
+    test('Fetching 0 or 1 rows(s) from the Pagila database without `columns()`', async () => {
+        // @@start-example@@ Find a single actor by id without `columns()`
```
fhur, line 60: This is not an easy title to search by. When composing this titles try to think of them as "google queries". 

In this case people would more likely search by something like Select all columns from the actors table

And with regards to the description, the important bit is to mention that if you omit the `select` all selectable fields will be selected by default.

packages/queries/src/getSelectableColumns.ts
```diff
@@ -0,0 +1,41 @@
+import { ColumnSchema, DbSchema, TableSchema } from './types/DbSchema';
```
fhur, line 1: Can you move this file to `src/schema`. We should only keep the most important files at the root.

packages/queries/src/getSelectableColumns.ts
```diff
@@ -0,0 +1,41 @@
...
+    return columnDef.properties.selectable.const;
+}
+
+/**
+ * Get all selectable columns of a table, as defined in the database schema.
+ *
+ * Example:
+ *
+ * ```ts
+ * const select = getSelectableColumns(schema, 'actor');
```
fhur, line 22: This could be a much better example if you showed me what `select` is

packages/queries/src/getSelectableColumns.ts
```diff
@@ -0,0 +1,41 @@
...
+ * ```
+ *
+ * @param schema The generated schema object for the database
+ * @param table The name of the table.
+ */
+
+export function getSelectableColumns<DB>(schema: DbSchema, table: Table<DB>) {
```
fhur, line 29: This is missing a function return type.

fhur:
Great job @jimezesinachi, feel free to merge once you've addressed the remaining comments. Also remember to squash 🙏

fhur (approved):

PR #10, opened 2024-05-22
feat: added better query providers

jimezesinachi:

## Why

Custom query providers currently have limited typing support. This PR improves that.

## What changed

QueryProvider and QueryExecutor

## Important


packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -14,17 +14,17 @@ type PgQueryResult = {
     [key: string]: any;
 };
 
-interface PgExecutorProps {
-    qpe?: QueryProviderExecutor;
+interface PgExecutorProps<DB> {
```
fhur, line 17: Why did you add the DB to the PgExecutor? It shouldn't be needed at this level.

packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -14,17 +14,17 @@ type PgQueryResult = {
...
+    qpe?: QueryProviderExecutor<DB>;
```
fhur, line 18: if it's for this type, just pass an `unknown`.

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
+
+        interface O {
+            in: Array<any>;
+        }
+
+        interface P {
+            '= any': any;
+        }
+
+        function isN(value: unknown): value is N {
```
fhur, line 28: what is this cryptic language @jimezesinachi  😂  isN, isO?

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
+        }
+
+        function isP(value: unknown): value is P {
+            return (
+                typeof (value as P)['= any'] === 'string' ||
+                typeof (value as P)['= any'] === 'number'
+            );
+        }
+
+        for (const [key, value] of Object.entries(where)) {
```
fhur, line 46: I can't tell if this covers all cases, but in essence this is what I was expecting 👍

packages/queries/src/types/QueryProviderInput.ts
```diff
@@ -0,0 +1,19 @@
+import { ColumnValue } from './ColumnValue';
+import { Table } from './Table';
+
+export type QueryProviderInput<
```
fhur, line 4: You should add some tests for this type

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
-    constructor(providers: QueryProvider[]) {
+export class QueryProviderExecutor<DB> implements QueryExecutor {
+    private providersByTable: Map<string, QueryProvider<DB, Table<DB>>>;
+
+    constructor(providers: QueryProvider<DB, Table<DB>>[]) {
         this.providersByTable = new Map(providers.map((p) => [p.table, p]));
     }
 
+    extract(where: Where<DB, Table<DB>>) {
+        const w: Record<string, any> = {};
```
fhur, line 16: I see you're going into a minimalist coding style now 😅 

Would be nice if you could expand `w` into something more readable 😉

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
+            );
+        }
+
+        for (const [key, value] of Object.entries(where)) {
+            if (isN(value)) {
+                w[key] = [value];
+            } else if (isO(value)) {
+                w[key] = [...value.in];
+            } else if (isP(value)) {
+                w[key] = [value['= any']];
```
fhur, line 52: this is not correct: this should be [...value['= any']]

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
-    private providersByTable: Map<string, QueryProvider>;
-    constructor(providers: QueryProvider[]) {
+export class QueryProviderExecutor<DB> implements QueryExecutor {
+    private providersByTable: Map<string, QueryProvider<DB, Table<DB>>>;
+
+    constructor(providers: QueryProvider<DB, Table<DB>>[]) {
         this.providersByTable = new Map(providers.map((p) => [p.table, p]));
     }
 
+    extract(where: Where<DB, Table<DB>>) {
```
fhur, line 15: I see you did some "divide and conquer" by splitting the function into it's own 👍 great call.

A few improvements we can make:
1. `extract` is not a great name. let's give it a much more explicit name. How about `convertWhereToQueryProviderInput`. It's not sexy, but it's not an API name, so it's totally fine to make it overly verbose (it literally has just 1 usage)
2. Move it outsude of the QueryProviderExecutor so you can actually write unit tests for it. This function is completely stateless, so making it actually stateless (i.e. just a plain old function) let's future maintainers quickly understand that the output only depends on the input.

packages/queries/src/types/QueryProviderInput.ts
fhur: Also add docs with examples.

packages/queries/src/types/QueryProviderInput.ts
```diff
@@ -0,0 +1,19 @@
+import { ColumnValue } from './ColumnValue';
```
fhur, line 1: I'm also thinking that this type shouldn't be in /queries, it should be in /backend as you would only ever use this type in a backend context.

packages/backend/src/execution/executors/QueryProviderExecutor.ts
jimezesinachi: Just placeholder names. Just wanted to confirm that the implementation idea was solid before fleshing out. Will fix all of them now.

packages/backend/src/execution/executors/QueryProviderExecutor.ts
jimezesinachi: Same thing I was thinking. Will add tests now to confirm 🚀

packages/backend/src/execution/executors/QueryProviderExecutor.ts
```diff
@@ -1,22 +1,73 @@
...
+
+        for (const [key, value] of Object.entries(where)) {
+            if (isN(value)) {
+                w[key] = [value];
+            } else if (isO(value)) {
+                w[key] = [...value.in];
+            } else if (isP(value)) {
+                w[key] = [value['= any']];
+            } else {
+                throw Error('Invalid where clause!');
```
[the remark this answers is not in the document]
jimezesinachi: Makes sense... on it 👍🏾

packages/backend/src/execution/executors/QueryProviderExecutor.ts
jimezesinachi: This! I actually had it that way before, but I wasn't sure and was going to ask you. Will update 👍🏾

packages/backend/src/query/convertWhereToQueryProviderInput.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, expect, test } from 'vitest';
+import { DB, from } from '../tests/generated';
+import { convertWhereToQueryProviderInput } from './convertWhereToQueryProviderInput';
+import { Table } from '@synthql/queries';
+
+describe('convertWhereToQueryProviderInput', () => {
```
fhur, line 6: This test is way too verbose, and is testing way too little cases jim. For example, you're creating an entire query, but not really using it. There's also a lot of code duplication between the test cases

I suggest you use the `test.each` used in other parts of the code. 

In essence you want two different types of cases:
- Case 1: wheres that should convert safely
- Case 2: wheres that should throw an error.

fhur:
I fixed the remaining issues, will proceed to merge now.

PR #11, opened 2024-05-27
feat added support for DOMAIN types

jimezesinachi:

## Why

Adding support for domain types

## What changed

Logic in `introspect/generate` and generated types and schemas in the packages (auto-generated)


packages/introspect/src/generate.ts
```diff
@@ -261,16 +262,27 @@ function createDomainJsonSchema(
...
+                `The ${domain.name} domain from the ${domain.schemaName} schema`,
         };
         return acc;
     }, empty);
 }
 
+function domainType(
+    domainInnerType: DomainDetails['innerType'],
+): JSONSchema['type'] {
+    if (domainInnerType === 'pg_catalog.int4') {
```
fhur, line 277: This looks roughly right to me, I guess we should iterate over the "createWellKnownDefs" type to ensure we're supporting all native PG types.

fhur (changes requested):
Good progress @jimezesinachi  👍 

Iterate over all known types and I think we will be done with this one.

packages/introspect/src/generate.ts
jimezesinachi: Makes sense. On it

fhur (approved):
LGTM 🚀

PR #12, opened 2024-05-27
feat: infer grouping ID based on schema

jimezesinachi:

## Why

We want to infer the grouping ID for a table when a user is using the query API

## What changed

A new function that extracts the column and passes it to the QueryBuilder, plus some updates to the tests that reflect this change

## Important

### DIFF
--- a/packages/queries/src/query.ts
+++ b/packages/queries/src/query.ts
@@ -5,6 +5,7 @@ import { Column } from './types/Column';
 import { Table } from './types/Table';
 import { DbSchema } from './types/DbSchema';
 import { getSelectableColumns } from './schema/getSelectableColumns';
+import { getPrimaryKey } from './schema/getPrimaryKey';
 
 export class QueryBuilder<
     DB,
@@ -332,7 +333,11 @@ export function query<DB>(schema: DbSchema<DB>) {
         from<TTable extends Table<DB>>(table: TTable) {
             type TKeys = Array<Column<DB, TTable>>;
 
+            type TPrimaryKey = Array<string>;
+
             const select = getSelectableColumns<DB>(schema, table);
+
+            const primaryKey = getPrimaryKey<DB>(schema, table);
 
             return new QueryBuilder<
                 DB,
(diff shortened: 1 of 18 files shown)


packages/queries/src/schema/getPrimaryKey.ts
```diff
@@ -0,0 +1,33 @@
...
+ * ```
+ *
+ * @param schema The generated schema object for the database
+ * @param table The name of the table
+ */
+
+export function getPrimaryKey<DB>(
+    schema: DbSchema<DB>,
+    table: Table<DB>,
+): string {
```
fhur, line 23: this isn't correct: this should return a `string[]` as multiple columns can be part of the primary key

packages/queries/src/query.ts
```diff
@@ -332,8 +333,12 @@ export function query<DB>(schema: DbSchema<DB>) {
         from<TTable extends Table<DB>>(table: TTable) {
             type TKeys = Array<Column<DB, TTable>>;
 
+            type TPrimaryKey = Array<string>;
```
fhur, line 336: you don't need to declare this type

packages/queries/src/schema/getPrimaryKey.ts
jimezesinachi: True that.. fixing now

packages/queries/src/query.ts
jimezesinachi: Yes, but the type being passed to `QueryBuilder` for `groupingId` (before my changes) was `['id']`, and not `string[]`. Still need the type passed to `QueryBuilder`, but I can pass it directly.

fhur: ah gotcha. Ok, just passing `string[]` is enough anyway. Alternatively you can probably do `typeof groupingId`

packages/queries/src/schema/getTableDefs.ts
```diff
@@ -22,10 +21,3 @@ export function isSelectableColumn(columnDef: ColumnSchema): boolean {
 export function isPrimaryKeyColumn(columnDef: ColumnSchema): boolean {
     return columnDef.properties.isPrimaryKey.const;
 }
-
-export function isValidColumnName<DB, TTable extends Table<DB>>(
```
fhur: 👍

fhur (approved):
LGTM!

PR #21, opened 2024-06-03
fix: added error handling

jimezesinachi:

## Why

Renamed/updated the SynthqlError class and used it in the Express handler

## What changed

The SynthqlError class and Express handler
 
## Important


packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -17,47 +17,112 @@ export function createExpressSynthqlHandler<T>(
...
+                query,
+                returnLastOnly,
+            );
 
-                res.end();
+            writeQueryResultToResponse(res, queryResult, returnLastOnly);
+        } catch (e) {
+            if (e instanceof SynthqlError) {
+                // If it's a known error, write it to the response
+                writeErrorToResponse(res, e);
```
fhur, line 35: you should await here

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -17,47 +17,112 @@ export function createExpressSynthqlHandler<T>(
...
-                    res.write('\n');
-                }
+            const queryResult = await tryExecuteQuery(
+                queryEngine,
+                query,
+                returnLastOnly,
+            );
 
-                res.end();
+            writeQueryResultToResponse(res, queryResult, returnLastOnly);
```
fhur, line 31: you should await here

fhur (changes requested):
Good progress, but not quite 😉 . Two big points stand out:

1. I think you're adding to many try/catch without understanding what they're actually doing. 
2. The idea of the SynthqlError is to encapsulate the error logic there.  Also worth mentioning: [here's the right way](https://chatgpt.com/c/6a52c9d3-5cf4-490e-b7b1-69bce438ddfa) to extend a class in js/ts. I left a few notes in the PR around this, let me know if it's not clear what you have to do :) 
3. You're missing `await` in quite a few places!
4. Lots of tests missing :)

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -17,47 +17,129 @@ export function createExpressSynthqlHandler<T>(
...
-                res.statusCode = 200;
-                res.setHeader('Content-Type', 'application/x-ndjson');
+                // Let another layer handle the error
+                throw e;
+            }
+        }
+    };
+}
+
+async function executeSynthqlRequest<T>(
```
fhur, line 41: Let's keep existing conventions: let's call T => DB

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
```diff
@@ -17,47 +17,129 @@ export function createExpressSynthqlHandler<T>(
...
+        });
+    }
+}
+
+async function tryExecuteQuery<T>(
+    queryEngine: QueryEngine<T>,
+    query: any,
+    returnLastOnly: boolean,
+) {
+    if (returnLastOnly) {
```
fhur, line 88: isn't this whole function just `queryEngine.execute(query, {returnLastOnly})`? no need for an `if`

packages/backend/src/SynthqlError.ts
```diff
@@ -0,0 +1,139 @@
...
+
+        return new SynthqlError(error, type, message);
+    }
+
+    static createResponseStreamingError({
+        error,
+        query,
+    }: {
+        error: any;
+        query: any;
```
fhur, line 64: we shouldn't use the `any` type here

packages/backend/src/SynthqlError.ts
```diff
@@ -0,0 +1,139 @@
...
+    }): SynthqlError {
+        const type = 'ResponseStreamingError';
+
+        const lines = [
+            'Response streaming error!',
+            '',
+            'Failure while streaming back the results for the query:',
+            '',
+            query,
```
fhur, line 73: if query is an object this will serialize to `[Object]`, we should do JSON.strinfigy(query,null,2)

fhur (changes requested):
Added some comments. Looks good overall. We'll probably be able to merge today after some added test cases 🙌

packages/backend/src/util/handlers/createExpressSynthqlHandler.ts
jimezesinachi: It should be, but at the time I wrote this, I wasn't sure how passing a falsy value to returnLastOnly was being used inside. After looking inside yesterday, I can see why this is unnecessary. Will fix now

packages/backend/src/SynthqlError.ts
fhur: for `query`

fhur (approved):

PR #22, opened 2024-06-06
feat: added benchmark tests & workflow

jimezesinachi:

## Why

We want to include performance benchmarks to show how the perfomance of SynthQL is changing over time. This PR adds this functionality

## What changed

Added benchmark tests and a workflow that runs on every merge into master

## Important


.github/workflows/deploy.yml
```diff
@@ -23,33 +31,77 @@ jobs:
...
+              run: git restore .
+
+            - name: Switch branch to gh-pages branch
+              run: git switch gh-pages
+
+            - name: Commit updated build files to gh-pages branch
+              run: |
+                  cp -r packages/docs/build/ ./
+                  git add .
+                  git commit -m "build: updated build files" || echo "Ignored commit failure, proceeding"
```
fhur, line 59: why ignore the commit failure?

jimezesinachi: The `cp` command prior might not always have modified files, so we only need to proceed to the next step if there's no modifications to commit

.github/workflows/deploy.yml
```diff
@@ -23,33 +31,77 @@ jobs:
...
-                  path: packages/docs/build
+                  name: SynthQL query execution benchmarks
+                  tool: 'benchmarkjs'
+                  output-file-path: packages/backend/src/tests/benchmarks/generated/output.txt
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  skip-fetch-gh-pages: true
+                  benchmark-data-dir-path: 'benchmarks'
+                  auto-push: true
+                  comment-always: true
+                  alert-threshold: '200%'
```
fhur, line 73: let's lower this down to 25%

and fail on 50%

fhur (dismissed):
LGTM added one small comment

fhur (approved):

fhur (approved):

PR #23, opened 2024-06-11
feat: add `offset()` and `take()` to the `query` API

jimezesinachi:

## Why

Adding some more methods to the `query` API

## What changed

Changes to `query`, `executePlan` and `useSynthql`

## Important


packages/backend/src/execution/execution/executePlan.ts
```diff
@@ -26,6 +26,8 @@ export function executePlan(
                 defaultSchema,
             );
 
+            if (parentNode) query.limit = undefined;
```
fhur, line 29: definitely add a comment, and always, always wrap in braces.

packages/backend/src/query/resolveReferences.ts
```diff
@@ -42,7 +42,8 @@ export function resolveReferences(
         ...query,
         where: resolveReferencesInWhere(query.where, refContext, defaultSchema),
         // Remove the limit, since we don't know how many rows the references will resolve to
-        limit: undefined,
+        // Commented out as this breaks limits in non-nested queries
```
fhur, line 45: Remove these lines, we don't need this logic here

fhur (changes requested):
Nice changes @jimezesinachi 👍 added a few small comments

fhur (approved):

PR #24, opened 2024-06-11
fix: connection error messages & fix: not handling `null` ref values

jimezesinachi:

## Why

Adding a DB connection error handler & fixed the error we get when the value referenced in an `include` is `null`

## What changed

Changes to the `SynthqlError`, `PgExecutor`, and `getIn`

## Important


packages/backend/src/util/tree/getIn.ts
```diff
@@ -49,19 +49,25 @@ function getNext(
     if (path.length === 0) {
         return [tree];
     }
-    if (tree === undefined) {
+
+    if (tree === undefined || tree === null) {
```
fhur, line 53: are you sure this is enough to fix it?

fhur: were you able to reproduce the issue?

packages/backend/src/execution/executors/PgExecutor/PgExecutor.test.ts
```diff
@@ -91,19 +91,17 @@ describe('PgExecutor', () => {
...
         .columns('film_id', 'title')
-        .where({ film_id: 1 })
         .include({
-            original_language_id: og_language,
+            original_language_id,
         })
-        .one();
+        .take(2);
 
     it('should', async () => {
```
fhur, line 106: can you add a nicer name?

packages/backend/src/execution/executors/PgExecutor/PgExecutor.test.ts
```diff
@@ -91,19 +91,17 @@ describe('PgExecutor', () => {
 
     const film_with_included_column_selected = from('film')
         .columns('film_id', 'title', 'original_language_id')
-        .where({ film_id: 1 })
         .include({
-            original_language_id: og_language,
+            original_language_id,
         })
-        .one();
+        .take(2);
```
fhur, line 97: nice!

packages/backend/src/execution/executors/PgExecutor/PgExecutor.test.ts
```diff
@@ -84,4 +81,49 @@ describe('PgExecutor', () => {
...
+        .columns('film_id', 'title')
+        .where({ film_id: 1 })
+        .include({
+            original_language_id: og_language,
+        })
+        .one();
+
+    it('should', async () => {
+        const result = await executor.execute(
+            film_with_included_column_selected,
```
fhur, line 108: also we shuoldn't mix the casing for these var names

PR #26, opened 2024-06-18
feat: basic property-based testing

jimezesinachi:

## Why

Added property-based regression tests to the `backend` package

## What changed

New test files and data extraction functions

## Important


packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
```diff
@@ -0,0 +1,109 @@
...
+import {
+    AnySchema,
+    getSelectableColumnsFromGenericDbSchema,
+    getWhereableColumnsFromGenericDbSchema,
+    getTableNamesFromGenericDbSchema,
+} from '@synthql/queries';
+
+type Cardinality = 'one' | 'maybe' | 'many';
+
+export function generateQueryArbitrary(
```
fhur, line 11: This needs some documentation

packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
```diff
@@ -0,0 +1,109 @@
...
+    schema: AnySchema,
+    allValuesMap: Map<string, Array<any>>,
+    cardinality: Cardinality,
+) {
+    return fc
+        .constantFrom(...getTableNamesFromGenericDbSchema(schema))
+        .chain((table) =>
```
fhur, line 18: Let's split this function up into more granular, easier to manage arbitraries. Let's also use a naming convention to easily identify them:

For example:

1. `function arbitraryTablename(schema: AnySchema)`
2. `arbitrarySelect(schema, table)`
3. `arbitraryWhere(...)`
4. `arbitraryLimit(...)`
5. `arbitraryCardinality(...)`

packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
```diff
@@ -0,0 +1,109 @@
...
+    column: string,
+): fc.Arbitrary<unknown> {
+    const tableValues = allValuesMap.get(table);
+
+    const columnValues = tableValues?.map((row) => {
+        const value = row[column];
+
+        return value;
+
+        // if (value instanceof Date) {
```
fhur, line 93: These should be removed

packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
fhur: This also means using the `arbitrary` prefix to name the top-level arbitraries like `arbitraryQuery` and `arbitraryEmptyQuery`

packages/backend/src/tests/propertyBased/properties/executeQuery.ts
```diff
@@ -0,0 +1,34 @@
+import { Pool } from 'pg';
+import { QueryEngine, collectLast } from '../../..';
+import { AnySchema, getTableNamesFromGenericDbSchema } from '@synthql/queries';
+
+export async function executeQuery<DB>(
+    queryEngine: QueryEngine<DB>,
+    query: any,
```
fhur, line 7: Don't be so liberal in your use of `any`s @jimezesinachi

fhur: `any`s are really dangerous!

packages/backend/src/tests/propertyBased/properties/executeQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+): Promise<any> {
+    const queryResult = await collectLast(
```
fhur, line 9: Maybe we should move this function to the `QueryEngine` itself. It's very often that you _just_ need the last value.

What do you think could be a good name for it? (maybe `executeAndWait`)?

packages/backend/src/tests/propertyBased/properties/executeQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+        queryEngine.execute(query, {
+            returnLastOnly: true,
+        }),
+    );
+
+    return queryResult;
+}
+
+export async function getTableValues(pool: Pool, schema: AnySchema) {
```
fhur, line 18: A more descriptive name would be `getValuesByTableName`

packages/queries/src/schema/getSelectableColumns.ts
```diff
@@ -25,18 +33,58 @@ type SelectableColumnsType = Record<string, true>;
...
+    for (const [columnName, columnDef] of getColumnNamesAndDefs<DB>(tableDef)) {
         if (isSelectableColumn(columnDef)) {
             select[columnName] = true;
         }
     }
 
     return select;
 }
+
+export function getSelectableColumnsFromGenericDbSchema(
```
fhur, line 52: it's now a bit weird that getSelectableColumns.ts is storing all these different functions.

My preference is to have one logical unit per file. This is not always possible, but I think it is in this case.

packages/queries/src/schema/getTableDefs.ts
```diff
@@ -1,23 +1,107 @@
...
+    AnyColumnDef,
+    AnySchema,
+    AnyTableDef,
+    ColumnDef,
+    Schema,
+    TableDef,
+} from '../types/Schema';
 import { Table } from '../types/Table';
 
+export function getTableNamesFromGenericDbSchema(schema: AnySchema): string[] {
```
fhur, line 12: same here: getTableDefs is now housing several different concepts

packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
```diff
@@ -0,0 +1,109 @@
...
+                    ),
+                limit: fc.integer({
+                    min: 1,
+                }),
+                cardinality: fc.constant(cardinality),
+            }),
+        );
+}
+
+export function generateEmptyQueryArbitrary(
```
fhur, line 67: Maybe you misunderstood me when I was taking about empty queries:

I actually meant that if in your db you have a column ID with values 1,2,3 and you search for select * from table where id = $1 and you know that $1 is NOT in the set 1,2,3 then that query should return:

1. many: []
2. one: throw
3. maybe: null

packages/backend/src/tests/propertyBased/arbitraries/cardinality.ts
fhur: Ideally also try to keep one arbitrary per file. This makes the files easier to read and reason about.

packages/backend/src/tests/propertyBased/arbitraries/arbitraries.ts
fhur, line 1: This file should also be split into separate files. One per exported arbitrary.

packages/backend/src/tests/propertyBased/arbitraries/arbitraryLimit.ts
```diff
@@ -0,0 +1,8 @@
+import { fc } from '@fast-check/vitest';
+
+export function arbitraryLimit(): fc.Arbitrary<number> {
+    return fc.integer({
+        min: 1,
+        max: 32767,
```
fhur, line 6: how did you get to this number?

Also the min should actually be zero, not one

packages/backend/src/tests/propertyBased/arbitraries/arbitraryQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+import { arbitrarySelect } from './arbitrarySelect';
+
+interface ArbitraryQuery<DB> {
+    schema: Schema<DB>;
+    allTablesRowsMap: AllTablesRowsMap;
+    cardinality: Cardinality;
+    validWhere: boolean;
+}
+
+export function generateArbitraryQuery<DB>({
```
fhur, line 18: rename to `arbitraryQuery`: pay attention to the little things @jimezesinachi. They don't matter by themselves but they are a good measure to see how well you're paying attention to the code that you're writing

packages/backend/src/tests/propertyBased/arbitraries/rbitraryWhere.ts
```diff
@@ -0,0 +1,80 @@
...
+}: {
+    schema: Schema<DB>;
+    allTablesRowsMap: AllTablesRowsMap;
+    tableName: string;
+    validWhere: boolean;
+}): fc.Arbitrary<Where<AnyDb, string>> {
+    return fc
+        .constantFrom(...getTableWhereableColumns(schema, tableName))
+        .chain((columnName) => {
+            // TODO: We should remove this check once we resolve the `timestampz` issue
```
fhur: You should add a description of the timestampz issue, we will certainly forget it otherwise.

In general try to make comments self contained. This way you can understand them without the need for additional context.

packages/backend/src/tests/propertyBased/properties/cardinalityMany.test.ts
```diff
@@ -1,62 +1,44 @@
...
     it.prop([validWhereArbitraryQuery], { verbose: 2 })(
         'Valid where query should return possibly empty array',
         async (query) => {
-            const queryResult = await executeQuery<DB>(queryEngine, query);
+            const queryResult = await executeQuery(queryEngine, query);
 
             expect(Array.isArray(queryResult)).toEqual(true);
 
-            expect(queryResult.length).toBeLessThanOrEqual(query.limit);
+            expect(queryResult.length).toBeLessThanOrEqual(Number(query.limit));
```
fhur: why are you wrapping query.limit in a Number? this shouldn't be needed.

packages/backend/src/tests/propertyBased/arbitraries/rbitraryWhere.ts
```diff
@@ -0,0 +1,80 @@
...
+                        depthIdentifier: '0',
+                        minKeys: 1,
+                        maxKeys: 1,
+                    },
+                );
+            }
+        });
+}
+
+function checkIfTimestampzColumn({
```
fhur: I'm a bit confused about the purpose of this function.

If it is to figure out if a column has the type `timestampz` this is not a good/reliable check. You should just get the PG type from the schema. This should be really easy to add to the schema.

packages/backend/src/tests/propertyBased/arbitraries/rbitraryWhere.ts
```diff
@@ -0,0 +1,80 @@
+import { fc } from '@fast-check/vitest';
```
fhur: this file has a typo in the name

packages/backend/src/tests/propertyBased/arbitraries/rbitraryWhere.ts
```diff
@@ -0,0 +1,80 @@
...
+        .chain((columnName) => {
+            // TODO: We should remove this check once we resolve the `timestampz` issue
+            if (
+                checkIfTimestampzColumn({
+                    allTablesRowsMap,
+                    tableName,
+                    columnName,
+                })
+            ) {
+                return arbitraryWhere({
```
fhur: why is arbitraryWhere invoking arbitraryWhere recursively?

packages/backend/src/tests/propertyBased/arbitraries/arbitraryWhereValue.ts
```diff
@@ -0,0 +1,86 @@
...
+export function arbitraryWhereValue({
+    allTablesRowsMap,
+    tableName,
+    columnName,
+    validWhere,
+}: {
+    allTablesRowsMap: AllTablesRowsMap;
+    tableName: string;
+    columnName: string;
+    validWhere: boolean;
```
fhur, line 13: What does `valid` mean in this context?

packages/backend/src/tests/propertyBased/arbitraries/rbitraryWhere.ts
jimezesinachi: This is fixed. Some of the review comments are appearing on outdated versions of the files.

packages/backend/src/tests/propertyBased/properties/cardinalityMany.test.ts
```diff
@@ -33,6 +33,12 @@ describe('cardinalityMany', async () => {
             expect(Array.isArray(result)).toEqual(true);
 
             expect(result.length).toBeLessThanOrEqual(Number(query.limit));
+
+            for (const item of result) {
+                for (const column of Object.keys(query.select)) {
```
fhur, line 38: This is not exactly true.

In theory you could have a query with `select: {id:false}` so just the prescence of the key is not enough.

A more precise test should be

```ts
const expectedKeys = Object.entries(query.select).flatMap(([key,selected]) => {
  return selected ? [key] : []
});

for (const item of result) {
  const actualKeys = Object.keys(item);
  expect(actualKeys).to.containSubset(expectedKeys)
}
```

This will also give a nicer error message if something breaks.

packages/backend/src/tests/propertyBased/properties/cardinalityMaybe.test.ts
```diff
@@ -33,6 +33,10 @@ describe('cardinalityMaybe', async () => {
             expect(result).toBeTypeOf('object');
 
             expect(Array.isArray(result)).toEqual(false);
+
+            for (const column of Object.keys(query.select)) {
+                expect(Object.keys(result)).toContain(column);
```
fhur, line 38: apply a similar change to the one suggested above

packages/backend/src/tests/propertyBased/properties/cardinalityOne.test.ts
```diff
@@ -36,6 +36,10 @@ describe('cardinalityOne', async () => {
             expect(Array.isArray(result)).toEqual(false);
 
             expect(result).not.toBeNull();
+
+            for (const column of Object.keys(query.select)) {
```
fhur, line 40: same here

packages/queries/src/index.ts
```diff
@@ -1,7 +1,26 @@
 export { col } from './col';
+export { getTablePrimaryKeyColumns } from './schema/getTablePrimaryKeyColumns';
```
fhur, line 2: Jim none of these functions belong here: 

These should not be part of the public API of SynthQL which we should try to keep minimal.

Am I missing something? Is there a reason why you placed them here?

packages/queries/src/schema/isEnumColumn.ts
```diff
@@ -0,0 +1,9 @@
+import { ColumnDef } from '../types/Schema';
+
+export function isEnumColumn(columnDef: ColumnDef): boolean {
+    if (columnDef.properties.type.enum) {
```
fhur, line 4: no need for an if statement. Just `return columnDef.properties.type.enum` is enough

packages/react/src/test/generated/schema.ts
```diff
@@ -344,8 +353,10 @@ export const schema = {
                                 'Column details:\n\n- Schema: public\n- Table: customer\n- Column: last_update\n- PG type: pg_catalog.timestamptz\n- PG kind: base\n- Nullable: true\n- Generated: NEVER',
                             properties: {
                                 type: {
+                                    id: 'pg_catalog.timestamptz',
                                     type: 'string',
```
fhur, line 357: wait: why is this `string`?

packages/backend/src/tests/propertyBased/arbitraries/arbitraryLimit.ts
```diff
@@ -0,0 +1,8 @@
+import { fc } from '@fast-check/vitest';
+
+export function arbitraryLimit(): fc.Arbitrary<number> {
+    return fc.integer({
+        min: 1,
+        max: 2147483647,
```
fhur, line 6: Always add a comment when using magic constants, we will forget where we got this from

packages/backend/src/tests/generated/db.ts
```diff
@@ -36,7 +36,7 @@ export type PgCatalogInt2 = number;
 /**
  * A PG numeric
  */
-export type PgCatalogNumeric = number;
+export type PgCatalogNumeric = string;
```
fhur, line 39: Can you add a comment to the JSON schema types explaining why this is a string so users understand what's going on (i.e. precision)

packages/backend/src/tests/propertyBased/arbitraries/arbitraryTableName.ts
```diff
@@ -0,0 +1,5 @@
+import { fc } from '@fast-check/vitest';
+
+export function arbitraryTableName(tableName: string): fc.Arbitrary<string> {
```
fhur, line 3: This is kind-of a useless abstraction

arbitraryTableName(schema) could be useful, but this is essentially just a layer of indirection

packages/backend/src/tests/propertyBased/arbitraries/arbitraryWhere.ts
```diff
@@ -0,0 +1,55 @@
...
+                // TODO: We should remove this check once we resolve the `date-time` issue
+                // When there's a mismatch between the timezone of the data stored in the
+                // database and the timezone of the machine that is running the database,
+                // the value returned from the database is adjusted to match the timezone
+                // of the server. But this means when we try to find rows matching the data
+                // received, we don't get the matching rows returned. So for now, we're using
+                // the logic below to exempt columns that of the timestampz type from being
+                // used in this property test
+
+                !checkIfDateTimeColumn({
```
fhur, line 32: do we still need this? I thought this was fixed?

packages/backend/src/tests/propertyBased/arbitraries/arbitraryWhereValue.ts
```diff
@@ -0,0 +1,114 @@
...
+
+            // Return an arbitrary value that matches two conditions:
+            // 1. It is of the type of the PG column gotten from the schema
+            // 2. It currently does not exist in the column values array
+
+            // The matching arms are presently incomplete
+            // We need to include more infomration in the schema
+            // to make them work better
+
+            if (columnPgType === 'pg_catalog.int2') {
```
fhur, line 51: These should all be arbitraries living in their own files. 

arbitraryInt2, arbitraryInt4, etc.

packages/backend/src/tests/propertyBased/arbitraries/arbitraryWhereValue.ts
```diff
@@ -0,0 +1,114 @@
...
+                return fc
+                    .integer({
+                        min: 1,
+                        max: 32767,
+                    })
+                    .filter((value) => !columnValuesFromSet.includes(value));
+            } else if (columnPgType === 'pg_catalog.int4') {
+                return fc
+                    .integer({
+                        min: 1,
```
fhur, line 61: you should be getting these bounds from the underlying JSON schema to also test that the schema is correct

fhur: same for most of the numeric schemas

packages/backend/src/tests/propertyBased/arbitraries/arbitraryWhereValue.ts
fhur: or see my comment below for another suggestion on how to clean this up a bit

fhur (commented):
Nice progress @jimezesinachi . Left a few comments. We're close to the finish line.

packages/queries/src/schema/isEnumColumn.ts
jimezesinachi: Enum is possibly `undefined`

fhur (approved):

PR #28, opened 2024-07-12
feat: update `Schema` type and schema file generator

jimezesinachi:

## Why

We're adding more type-safety to the generated schema

## What changed

The schema generator function, its API, and the CLI generator function, and its API

## Important


packages/introspect/src/e2e.test.ts
```diff
@@ -22,7 +22,9 @@ describe('e2e', () => {
                     parser: 'typescript',
                     ...prettierOptions,
                 }),
-            SECRET_INTERNALS_DO_NOT_USE_queriesImportLocation: '../query',
+            SECRET_INTERNALS_DO_NOT_USE_queryBuilderImportLocation: '../query',
```
fhur, line 25: You don't need two secret params. You can just have a single one that points to the "index" file and from there export everything you need.

packages/introspect/src/generate.ts
```diff
@@ -177,10 +189,17 @@ function createTableJsonSchema(table: TableDetails): JSONSchema {
 function createRootJsonSchema(
     schemas: Record<string, Schema>,
     { defaultSchema }: { defaultSchema: string },
```
fhur, line 191: don't add a third parameter, instead add "incldueTables" to the second "props-object" that currently only contains defaultSchema.

fhur (approved):
Added a few extra fixed and merging

PR #29, opened 2024-07-16
feat: add CLI config file

jimezesinachi:

## Why

## What changed

Brief summary of the changes.

## Important

reviewers.


packages/cli/src/commands/generate/index.ts
```diff
@@ -15,14 +17,44 @@ export const generateSchema = async ({
...
+    if (schemaDefOverrides === undefined) {
+        try {
+            return await generate({
+                defaultSchema,
+                connectionString,
+                includeSchemas: schemas,
+                includeTables: tables,
+                outDir: path.resolve(path.join(process.cwd(), out)),
+            });
+        } catch (error) {
```
fhur, line 31: Don't catch an error just to log it. This isn't really doing anything. Just let the error bubble up.

fhur: Same for the other try/catch blocks in this file.

packages/cli/src/commands/generate/index.ts
```diff
@@ -15,14 +17,44 @@ export const generateSchema = async ({
...
+            `${JSON.stringify(schemaDefOverrides, null, 2)}`,
+        ];
 
-    return result;
+        throw new Error(lines.join('\n'));
+    }
 };
+
+function isValidSchemaDefOverrides(value: any): value is SchemaDefOverrides {
+    return value satisfies SchemaDefOverrides;
```
fhur, line 59: this function is not doing anything at runtime. This function is literally equivalent to 
```ts
function isValidSchemaDefOverrides(x) {return x}
```

packages/cli/src/cli.ts
```diff
@@ -47,22 +47,35 @@ export function cli({
...
                     schemas: argv.schemas,
                     tables: argv.tables,
+                    schemaDefOverrides: argv.schemaDefOverrides,
                 });
-                const tables = Object.keys(schema.properties ?? {});
-                console.warn(`Generated schema with ${tables.length} table(s)`);
-                for (const table of tables) {
-                    console.warn(`- ${table}`);
+
+                if (result) {
```
fhur, line 69: This if isn't needed: this function should always return something or throw: see my point further down about not handling the try/catches if you can't actually handle them.

packages/cli/src/commands/generate/index.ts
```diff
@@ -15,14 +17,44 @@ export const generateSchema = async ({
...
+                includeSchemas: schemas,
+                includeTables: tables,
+                outDir: path.resolve(path.join(process.cwd(), out)),
+            });
+        } catch (error) {
+            console.log('Error: ', JSON.stringify(error, null, 2));
+        }
+    } else if (isValidSchemaDefOverrides(schemaDefOverrides)) {
+        try {
+            return await generate({
```
fhur, line 36: you don't need to calls to `generate`, you should call it just once and always pass the overrides, which can possibly be empty.

packages/introspect/src/generate.ts
```diff
@@ -136,13 +154,25 @@ function createTableJsonSchema(table: TableDetails): JSONSchema {
...
                 ].join('\n'),
             properties: {
-                type: { $ref: `#/$defs/${createTypeDefId(column.type)}` },
-                // a constant value of true
-                selectable: { type: 'boolean', const: true },
-                includable: { type: 'boolean', const: true },
-                whereable: { type: 'boolean', const: true },
-                nullable: { type: 'boolean', const: column.isNullable },
-                isPrimaryKey: { type: 'boolean', const: column.isPrimaryKey },
+                type: columnSchemaOverrides?.type
```
fhur, line 157: Did you try doing this? 

```ts
           {
                type: { $ref: `#/$defs/${createTypeDefId(column.type)}` },
                // a constant value of true
                selectable: { type: 'boolean', const: true },
                includable: { type: 'boolean', const: true },
                whereable: { type: 'boolean', const: true },
                nullable: { type: 'boolean', const: column.isNullable },
                isPrimaryKey: { type: 'boolean', const: column.isPrimaryKey },
                ...overrides
            }

File: packages/introspect/src/generate.ts [hunk 6]
```diff
@@ -244,12 +276,21 @@ function fullTableName(table: TableDetails, defaultSchema: string) {
...
     tables: TableDetails[],
     defaultSchema: string,
+    schemaDefOverrides: SchemaDefOverrides | undefined,
 ): Record<string, JSONSchema> {
     const empty: Record<string, JSONSchema> = {};
 
     return tables.reduce((acc, table) => {
-        acc[createTableDefId(table, defaultSchema)] =
-            createTableJsonSchema(table);
+        const tableSchemaOverrides = schemaDefOverrides
```
fhur, line 284: you don't need a call to `.includes`. This should be a simple call to `schemaDefOverride[table]`

packages/introspect/src/generate.ts
```diff
@@ -186,12 +216,14 @@ function createRootJsonSchema(
     {
         defaultSchema,
         includeTables,
+        schemaDefOverrides,
     }: {
         defaultSchema: string;
         includeTables: string[];
+        schemaDefOverrides: SchemaDefOverrides | undefined;
```
fhur, line 223: instead of passing undefined around all the time, you can just `?? {}` at the top level.

fhur (changes requested):
Good progress jim 👍 

A few things can be simplified but otherwise looks pretty much what I expected.

packages/cli/src/cli.ts
```diff
@@ -47,22 +47,35 @@ export function cli({
                         description:
                             'List of tables to be included in the generation',
                     })
-                    .array('tables');
+                    .array('tables')
+                    .config('configFile')
+                    .parserConfiguration({
```
fhur, line 52: what is this for?

packages/cli/src/tests/synthql.config.ts
```diff
@@ -0,0 +1,26 @@
+import { CliConfig } from '../types/CliConfig';
+
+export const config: CliConfig = {
+    out: 'src/generated/synthql',
+    defaultSchema: 'public',
+    schemas: ['public'],
+    tables: ['actor', 'customer'],
+    schemaDefOverrides: {
```
fhur, line 8: did you check if the overrides will work when you have two tables with the same name defined in different schemas?

fhur: To fix I suggest you require schema.table notation (e.g. public.actor). Its slightly more verbose, but more explicit.

packages/introspect/src/generate.ts
fhur: Much simpler, and scalable implementation.

packages/cli/src/cli.ts
```diff
@@ -47,22 +47,35 @@ export function cli({
...
-                const tables = Object.keys(schema.properties ?? {});
-                console.warn(`Generated schema with ${tables.length} table(s)`);
-                for (const table of tables) {
-                    console.warn(`- ${table}`);
+
+                if (result) {
+                    const tables = Object.keys(result.schema.properties ?? {});
+
+                    console.warn(
+                        `Generated schema with ${tables.length} table(s)`,
```
fhur, line 73: You should provide some basic feedback to the user in case the overrides are passed but not actually applied to anything.

packages/cli/src/cli.ts
fhur: Ping @jimezesinachi

packages/cli/src/cli.ts
```diff
@@ -47,20 +47,32 @@ export function cli({
...
                     connectionString: argv.connectionString,
                     out: argv.out,
                     defaultSchema: argv.defaultSchema,
                     schemas: argv.schemas,
                     tables: argv.tables,
+                    schemaDefOverrides: argv.schemaDefOverrides,
                 });
-                const tables = Object.keys(schema.properties ?? {});
+
+                // TODO: We want to check if there were any overrides passed
```
fhur, line 69: This should be implemented in this MR. It's a very simple check.

Just check if the generated tables include the overrides.

jimezesinachi: Done now.. I added the comment so that I'll remember to add it this morning

packages/cli/src/helpers/getUnappliedSchemaOverridesTableNames.ts
```diff
@@ -0,0 +1,29 @@
+import { isObject } from './isValidSchemaDefOverrides';
+
+export function getUnappliedSchemaOverridesTableNames(
```
fhur, line 3: remember what I said about co-location: this function is in a "helpers" folder, but it will only ever be used in a single place. So better to co-locate it.

Also in general, try to avoid "helper" or "util" folders. Some things are actually helper/utils but quite often it's just a lazy way to drop unrelated code. So helper folders become a noisy bag of unrelated functionality over time.

packages/cli/src/tests/synthql.config.json
```diff
@@ -16,6 +16,22 @@
             "last_name": {
                 "includable": { "type": "boolean", "const": false }
             }
+        },
+        "public.niggas": {
```
fhur: I don't know jim... I feel like we shouldn't use this word in an open source repo 😅

packages/cli/src/helpers/getUnappliedSchemaOverridesTableNames.ts
```diff
@@ -0,0 +1,29 @@
...
+    tables: string[],
+    schemaDefOverrides: unknown,
```
fhur, line 5: You shuold first validate this type, using the JSON schema approach we mentioned earlier and then pass the right type here.

packages/cli/src/helpers/isValidSchemaDefOverrides.ts
```diff
@@ -0,0 +1,113 @@
...
+};
+
+const columnDefPropertyKeys = Object.keys(columnDefProperties);
+
+export function isValidSchemaDefOverrides(
+    value: unknown,
+): value is SchemaDefOverrides | undefined {
+    if (isObject(value)) {
+        for (const [tableDefName, tableDefOverride] of Object.entries(value)) {
+            if (containsSingleDot(tableDefName)) {
```
fhur, line 19: Jim why didn't you go with the JSON schema approach?

https://discord.com/channels/1222945509266817055/1222945509740642317/1263412303445889044

fhur: Also I feel like you haven't really "grokked" the idea of returning early

fhur: Here is a sample schema btw:

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "properties": {
        "tables": {
            "type": "array",
            "items": {
                "type": "string"
            }
        },
        "schemas": {
            "type": "array",
            "items": {
                "type": "string"
            }
        },
        "defaultSchema": {
            "type": "string",
        },
        "schemaDefOverrides": {....}
    },
    "required": ["tables", "schemaDefOverrides"]
}
```

packages/cli/src/tests/synthql.config.json
jimezesinachi: I agree.. removed already

packages/cli/src/commands/generate/index.ts
```diff
@@ -1,12 +1,15 @@
-import { generate } from '@synthql/introspect';
 import path from 'path';
+import { generate } from '@synthql/introspect';
+import { isValidSchemaDefOverrides } from '../../validators/isValidSchemaDefOverrides';
+import { SchemaDefOverrides } from '@synthql/queries';
```
fhur, line 4: SchemaDefOverrides is a CLI-only construct. It should therefore be defined in the CLI package.

packages/cli/src/commands/generate/index.ts
```diff
@@ -15,14 +18,14 @@ export const generateSchema = async ({
...
 }: GenerateSchemaOptions) => {
-    const result = await generate({
+    return await generate({
         defaultSchema,
         connectionString,
         includeSchemas: schemas,
         includeTables: tables,
-        outDir: out,
+        schemaDefOverrides: schemaDefOverrides,
+        outDir: path.resolve(path.join(process.cwd(), out)),
```
fhur, line 29: jim you're not paying attention: this resolve/join/cwd pattern is buggy, it doesn't respect absolute paths.

Consider what happens when you use the absolute path "/tmp/foo.ts", it will be converted into ./tmp/foo.ts, which is incorrect.

I've mentioned this error in the past, please don't introduce this bug again.

packages/cli/src/commands/generate/index.ts
jimezesinachi: That is not correct 😅... It is also used in the introspect package

packages/cli/src/validators/schemas.ts
```diff
@@ -0,0 +1,341 @@
+import { JSONSchemaType } from 'ajv';
+import { CliConfig } from '../types/CliConfig';
+
+export const configFileSchema: JSONSchemaType<Partial<CliConfig>> = {
```
fhur, line 4: there is a LOT of duplication in this file.

I would suggest splitting out a `tableSchema` and adding a function to make some fields nullable, similar to typescripts `Nullable<T>` type.

packages/cli/src/validators/schemas.ts
```diff
@@ -0,0 +1,341 @@
...
+            nullable: true,
+            required: [],
+            additionalProperties: false,
+        },
+    },
+    required: [],
+    additionalProperties: false,
+};
+
+export const cliArgvSchema: JSONSchemaType<CliConfig> = {
```
fhur, line 175: I'm not sure why this is needed 🤔 doesn't `yargs` provide type-safety already?

packages/queries/src/types/Schema.ts
```diff
@@ -47,38 +47,46 @@ export interface ColumnDef {
...
-        };
-        nullable: {
-            type: string;
-            const: boolean;
-        };
-        isPrimaryKey: {
-            type: string;
-            const: boolean;
-        };
+        type: ColumnDefTypeDef;
```
fhur, line 50: I'm not sure this is an improvement in readability over the previous version.

`{type:string, const:boolean}` is much easier to read than `ColumnDefAttributeDef`. Seems like a layer of indirection to me.

packages/introspect/src/generate.ts
```diff
@@ -116,10 +148,25 @@ function createTableDefId(type: TableDetails, defaultSchema: string) {
...
 }
 
-function createTableJsonSchema(table: TableDetails): JSONSchema {
+function createTableJsonSchema(
+    table: TableDetails,
+    tableDefOverrides: TableDefOverrides | undefined,
+): JSONSchema {
     const empty: Record<string, any> = {};
 
+    // Here, we want to check if there was overriding table schema def
```
fhur, line 157: I think you're missing an `applyOverrides` function that encapsulates this logic.

packages/cli/src/cli.ts
```diff
@@ -47,22 +50,76 @@ export function cli({
...
-                });
-                const tables = Object.keys(schema.properties ?? {});
-                console.warn(`Generated schema with ${tables.length} table(s)`);
-                for (const table of tables) {
-                    console.warn(`- ${table}`);
+                // We do this so that we lose the types from yargs but we gain
+                // complete type-safety after we pass the data through the validator
+                const data = { ...argv, connectionString: undefined } as any;
+
+                if (validateCliArgv(data)) {
```
fhur, line 81: Please use the return early pattern. This is much harder to read this way. It breaks being able to linearly scan the text.

packages/cli/src/cli.ts
```diff
@@ -47,22 +50,76 @@ export function cli({
...
-                    schemas: argv.schemas,
-                    tables: argv.tables,
-                });
-                const tables = Object.keys(schema.properties ?? {});
-                console.warn(`Generated schema with ${tables.length} table(s)`);
-                for (const table of tables) {
-                    console.warn(`- ${table}`);
+                // We do this so that we lose the types from yargs but we gain
+                // complete type-safety after we pass the data through the validator
+                const data = { ...argv, connectionString: undefined } as any;
```
fhur, line 79: Don't use `as any`. For cases like these use `as unknown`. 

`as any` is dangerous, `as unknown` is safe.
`as any` allows any compiler errors because it allows any type.
`as unknown` will force you to check for the type, so it's totally safe to use it.

packages/cli/src/commands/generate/index.ts
fhur: Yeah about that: we also need to refactor that. We shouldn't couple the  /introspect package to the CLI. I'm about to post a comment around that.

packages/cli/src/commands/generate/index.ts
```diff
@@ -15,14 +17,14 @@ export const generateSchema = async ({
...
+    schemaDefOverrides,
 }: GenerateSchemaOptions) => {
-    const result = await generate({
+    return await generate({
         defaultSchema,
         connectionString,
         includeSchemas: schemas,
         includeTables: tables,
-        outDir: out,
+        schemaDefOverrides: schemaDefOverrides,
```
fhur, line 27: SchemaDefOverrides is a `CLI`-only construct. It shouldn't "leak" into the /introspect package.

We currently only allow synthql.cofig.json, but in the near future we will also allow synthql.config.(js|ts), where we will allow users to maybe functions to override the schemas.

The important bit to take from this is that schemaDefOverrides as a concept is very much dependent on the shape of the synthql.config.json and we shouldn't leak that implementation detail into the /introspect pacakge.

As you also noticed, this also forces us to share a type between these two packages through the /queries package (which has nothing to do with this at all).

So what's the proper fix? Quite simple, actually. See the following pseudocode:

1. Add a property on the generate function that allows for passing an array of table overriders.
```ts
Array<{
    test: (table) => boolean,
    applyOverride: (oldTable) => NewTable
}>
```
These overriders can specify a `test` to determine on which tables they should apply to, and when the test is true, a function that applies the override.

2. Add a function that takes `synthql.config.json` format and creates an array of overriders 
```ts
const overrides = createTableOverrides(jsonConfig)
```

packages/cli/src/validators/schemas.ts
jimezesinachi: This is what I originally tried to do, but for type breaks whenever you assign a passed variable to the JSONSchemaType

packages/cli/src/validators/schemas.ts
jimezesinachi: It does, but because schemaDefOverrides is read from the config file, it's type is unknown to yargs, so we use that to give the values their correct type

packages/cli/src/commands/generate/index.ts
fhur: See https://github.com/synthql/SynthQL/pull/29#discussion_r1686162734

packages/cli/src/validators/schemas.ts
fhur: Ok fair, but that means we just need a schema for the schemaDefOverrides, not for all the other args, right?

jimezesinachi: That is correct... I will adjust it now 👍🏾

packages/queries/src/types/Schema.ts
jimezesinachi: Wasn't much for readablity as it was for reusability (i.e. deduplication)... but if it's severely affecting readability, as can switch it back

fhur: De-duplication between the fields?

jimezesinachi: Yes.. the selectable, whereable, includeable, nullable and isPrimaryKey fields are all assigned to an identical object: `{ type:string, const:boolean }`, which is what `ColumnDefAttributeDef` is equal to

packages/cli/src/e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+            // Verify that the passed attribute overrides were applied
+            const schemaFilePath = path.join(outDir, `schema.ts`);
+
+            const file = await import(schemaFilePath);
+            const schema = file.schema.properties;
+
+            const schemaDefOverrides = output.schemaDefOverrides as any;
+
+            for (const qualifiedTableName in schemaDefOverrides) {
```
fhur, line 107: wow what's happening here? what are you trying to assert?

packages/cli/src/e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+            // Verify that the option data is correct
+            expect(output.configFile).toEqual(configFile);
+
+            // Verify that the passed attribute overrides were applied
+            const schemaFilePath = path.join(outDir, `schema.ts`);
+
+            const file = await import(schemaFilePath);
+            const schema = file.schema.properties;
+
+            const schemaDefOverrides = output.schemaDefOverrides as any;
```
fhur, line 105: Please don't use `any`. Is there a reason why you need it here?

packages/cli/src/e2e.test.ts
fhur: Usually it's not a good idea to have too much logic inside your tests, as bugs can creep into the logic over time.

Can we simplify this logic? Why can't we just do a single big `expect`?

packages/cli/src/tests/synthql.dev.config.json
```diff
@@ -0,0 +1,179 @@
+{
```
fhur, line 1: what is the purpose of this file?

packages/cli/src/types/CliConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+     * The table definitions for any tables you want
+     * to be overriden during the schema generation
+     */
+    schemaDefOverrides?: SchemaDefOverrides;
+}
+export interface SchemaDefOverrides {
+    /**
+     * The column definitions for each table to be overriden
+     * The key for each table should be the qualified name for the table,
+     * in the form `${schemaName.tableName}` e.g `'public.users'`
```
fhur, line 36:
```suggestion
     * in the form `${schemaName}.${tableName}` e.g `'public.users'`
```

packages/cli/src/types/CliConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+     */
+    [x: string]: TableDefOverrides | undefined;
```
fhur, line 38: I would suggest we don't introduce `TableDefOverrides`. Also that we give `x` a better name.

```suggestion
    [table: string]: { [column: string]: Partial<ColumnDefProperties> }
```

packages/cli/src/types/CliConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+export interface TableDefOverrides {
```
fhur, line 40: See my suggestion above to remove this. As a general rule of thumb, try to keep the number of names you introduce into the codebase as small as possible.

Just because something has a name doesn't mean that the name is more descriptive than the thing being named.

For example `T | null` is much more readable and descriptive than `Option<T>`. There's two reasons for this:
1. `T | null` will be familiar to every single TS/JS dev in the world, unlike `Option<T>` which could be implemented as `T | null` or `T | undefined` or `T | null | undefined`
2. A name's usefulness increases in proportion to how often the name is used in the codebase. Assuming that Option<T> was implemented as `T | null | undefined` and this pattern was used everywhere in the codebase, then that probably OK. Now consider the `TableDedOverrides` which is used only once (maybe twice?) in the entire codebase. People that look at that name will always have to look at the implementation.

packages/cli/src/validators/createTableDefOverriderList.ts
```diff
@@ -0,0 +1,36 @@
...
+import { SchemaDefOverrides } from '../types/CliConfig';
+
+interface TableDefOverrider {
+    isValidQualifiedTableName: (qualifiedTableName: string) => boolean;
+    applyColumnDefOverride: (
+        columnName: string,
+    ) => Partial<ColumnDefProperties> | undefined;
+}
+
+type TableDefOverriderList = Array<TableDefOverrider>;
```
fhur, line 11: Another example of bad naming. The name is worse than the thing being named. This is just a layer of indirection.

packages/cli/src/validators/createTableDefOverriderList.ts
```diff
@@ -0,0 +1,36 @@
+import { ColumnDefProperties } from '@synthql/queries';
+import { SchemaDefOverrides } from '../types/CliConfig';
+
+interface TableDefOverrider {
+    isValidQualifiedTableName: (qualifiedTableName: string) => boolean;
+    applyColumnDefOverride: (
+        columnName: string,
+    ) => Partial<ColumnDefProperties> | undefined;
```
fhur, line 8: This shouldn't ever return `undefined`.

packages/cli/src/validators/createTableDefOverriderList.ts
```diff
@@ -0,0 +1,36 @@
+import { ColumnDefProperties } from '@synthql/queries';
+import { SchemaDefOverrides } from '../types/CliConfig';
+
+interface TableDefOverrider {
+    isValidQualifiedTableName: (qualifiedTableName: string) => boolean;
```
fhur, line 5: "valid" is the wrong adjective here. I would call this `test`, `predicate`, `isApplicable`.

packages/cli/src/validators/createTableDefOverriderList.ts
```diff
@@ -0,0 +1,36 @@
...
+    applyColumnDefOverride: (
```
fhur, line 6: This signature doesn't make sense. This should take an existing `ColumnDefProperties` as input and just transform it.

packages/cli/src/validators/getUnappliedSchemaOverridesTableNames.ts
```diff
@@ -0,0 +1,33 @@
+import { SchemaDefOverrides } from '../types/CliConfig';
+
+export function getUnappliedSchemaOverridesTableNames(
+    tables: string[],
+    defaultSchema: string,
+    schemaDefOverrides?: SchemaDefOverrides,
+) {
+    const unappliedSchemaOverrideTableKeys: string[] = [];
+
+    if (isObject(schemaDefOverrides)) {
```
fhur, line 10: you don't need isObject. Simply do 

```
if (!schemaDefOverrides) {
    return []
}
```

packages/cli/src/tests/synthql.dev.config.json
fhur: nvm, for the tests

packages/cli/src/e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+} from 'vitest';
+import * as prettier from 'prettier';
+import fs from 'fs';
+import path from 'path';
+import { configFileSchema } from './validators/schemas';
+import { cli } from './cli';
+import { randomUUID } from 'crypto';
+
+describe('e2e', () => {
+    const out = 'generated/synthql/e2e/tests';
```
fhur, line 18: if these are temporary files it would be better to write them to /tmp

packages/cli/src/e2e.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    vitest,
+} from 'vitest';
+import * as prettier from 'prettier';
+import fs from 'fs';
+import path from 'path';
+import { configFileSchema } from './validators/schemas';
+import { cli } from './cli';
+import { randomUUID } from 'crypto';
+
+describe('e2e', () => {
```
fhur, line 17: Conceptually, what's the difference between `src/e2e.test.ts` and `src/cli.test.ts`?

packages/cli/src/validators/getUnappliedSchemaOverridesTableNames.ts
fhur: In fact, here's a simplified version:

```
import { SchemaDefOverrides } from '../types/CliConfig';

export function findUnusedOverrides(
    existingTables: string[],
    defaultSchema: string,
    schemaOverrides?: SchemaDefOverrides,
): string[] {
    if (!Boolean(schemaOverrides)) {
        return [];
    }

    return Object.keys(schemaOverrides).flatMap((fullTableName) => {
        const [schemaName, tableName] = fullTableName.split('.');
        const isDefaultSchema = schemaName === defaultSchema;
        const tableToCheck = isDefaultSchema ? tableName : fullTableName;

        return !existingTables.includes(tableToCheck) ? [`- ${fullTableName}`] : [];
    });
}
```

packages/cli/src/validators/getUnappliedSchemaOverridesTableNames.ts
```diff
@@ -0,0 +1,33 @@
...
+        const tablesToOverride = Object.keys(schemaDefOverrides);
+
+        for (const tableKey of tablesToOverride) {
+            const [schemaName, tableName] = tableKey.split('.');
+
+            if (schemaName === defaultSchema) {
+                if (!tables.includes(tableName)) {
+                    unappliedSchemaOverrideTableKeys.push(`- ${tableKey}`);
```
fhur, line 18: It's really weird that this function returns an array of `- table`. It completely breaks the potential composability of this function.

I would suggest you remove the `-` part.

It's like mixing two concerns into one: the rendering (`- `) and the logic (finding unused overrides).

packages/cli/src/validators/getUnappliedSchemaOverridesTableNames.ts
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+        }
+    }
+
+    return unappliedSchemaOverrideTableKeys;
+}
+
+export function isObject(value: unknown): value is object {
```
fhur, line 31: Are we using this anywhere else? if not, we should remove it as it's not really needed in this function.

packages/cli/src/e2e.test.ts
jimezesinachi: `e2e.test.ts` is just where I'm adding any tests where we're testing the results (and not just the `CLI` function, like the tests in `cli.test.ts`) and/or are storing the side-effects, using it the same way you used it in the introspect package

packages/cli/src/tests/synthql.dev.config.json
jimezesinachi: It's a copy of the config schema I'm using to validate the version I'm using for the tests

packages/cli/src/e2e.test.ts
jimezesinachi: Indeed. Same reason why we needed the` ajv `typing. `schemaDefOverrides` is always `unknown`. I thought of using the validator function (`validateSchemaDefOverrides`)to validate the value, but that would be adding a dependency to the tests. The idea of using any here is that we want this test to fail with a reference error, the moment the data shape is not what it expects, without having to use a function from a dependency

packages/cli/src/validators/getUnappliedSchemaOverridesTableNames.ts
fhur: Yeah, for sure. there is essentially _zero_ cost to iterating what will in the worst case be a few 1000 items.

packages/cli/src/tests/synthql.dev.config.json
fhur: Can you add a comment somewhere (i.e. where you generate it) explaining what this is for

packages/cli/src/e2e.test.ts
jimezesinachi: I will look into simplifying it, but a big expect might not be possible because the objects don't have the same shape

packages/cli/src/e2e.test.ts
fhur: I think using the validator makes a lot of sense, specially given that this is an e2e test (we're testing the whole flow, including validation). Why would adding a dependency on it be a bad idea?

jimezesinachi: The way I was thinking about it, since we were using the validator to parse the data inside the CLI function, it seemed like testing the output of the validator process using the same validator... but if we're okay with that, then it makes sense to use it

packages/cli/src/types/CliConfig.ts
```diff
@@ -0,0 +1,53 @@
+import { ColumnDefProperties } from '@synthql/queries';
+export interface CliConfig {
+    /**
+     * Internal. DO NOT USE/SET this option.
+     * Specifies which draft of the JSON Schema
+     * standard the schema adheres to.
+     */
+    $schema?: string;
```
fhur, line 8: instead of string we should probably just spe

fhur (dismissed):

fhur (approved):

PR #30, opened 2024-07-26
feat: added prepending SQL queries

jimezesinachi:

## Why

We need the ability to run SQL queries before SynthQL queries

## What changed

Changes to `execute()`, `PgExecutor` and `QueryEngine`

## Important


packages/backend/src/QueryEngine.ts
```diff
@@ -13,17 +13,20 @@ import { SynthqlError } from './SynthqlError';
...
     schema?: string;
+    prependSql?: string;
     providers?: Array<QueryProvider<DB, Table<DB>>>;
     pool?: Pool;
 }
 
 export class QueryEngine<DB> {
     private pool: Pool;
     private schema: string;
+    private prependSql: string;
```
fhur, line 24: this shouldn't be global: it should be set on a per-request basis.

jimezesinachi: Fair enough, will adjust now 👍🏾

packages/backend/src/execution/types.d.ts
```diff
@@ -7,10 +7,12 @@ import { QueryNode } from '../query/createQueryTree';
 /**
  * # Execution Plan
  *
- * The execution plan tree describes the steps that need to be taken to execute a query,
- * as well as which strategies are taken to execute each query.
+ * The execution plan tree describes the steps that
```
fhur, line 10: jim what are all these formatting changes?

packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -43,20 +42,29 @@ export class PgExecutor implements QueryExecutor<PgQueryResult> {
...
+
         const { params, sql } = sqlBuilder.build();
 
         try {
             if (this.props.logging) {
                 console.log(format(sql, { language: 'postgresql' }));
             }
 
+            if (prependSql) {
+                await client.query(prependSql);
```
fhur, line 65: that's not right: we shouldn't open two SQL queries, we want to concatenate the queries so they are part of the same transaction.

fhur: So we want something like this:

```
begin
{prependSql};
{sql};
commit
```

fhur: So this would look like this for Luminovo's case:

```ts
const transformSql = (sql:string) => {
    return `begin set "my.tenant"=`someTenant`; ${sql}; commit;`
}
queryEngine.execute(query,{transformSql})
```

packages/backend/src/execution/types.d.ts
jimezesinachi: Just for readability.. was having to scroll horizontally considerably, and that was hindering me grasping what I was trying to read

packages/backend/src/QueryEngine.ts
```diff
@@ -46,6 +46,7 @@ export class QueryEngine<DB> {
         query: TQuery,
         opts?: {
             schema?: string;
+            transformSql?: (sql: string) => string;
```
fhur, line 49: Missing documentation and a unit test

fhur (changes requested):
A small docstring and a unit test and this should be good to go :)

packages/backend/src/execution/executors/PgExecutor/PgExecutor.test.ts
```diff
@@ -83,6 +83,27 @@ describe('PgExecutor', () => {
...
+    const q3 = from('actor')
+        .columns('actor_id', 'first_name', 'last_name')
+        .where({ actor_id: 1 })
+        .one();
+
+    it(`Actor table SynthQL query executes to 
+        expected result with SQL to prepend included`, async () => {
+        const result = await executor.execute(q3, {
+            ...executeProps,
+            prependSql: 'BEGIN',
```
fhur, line 95: this test doesn't make sense 🤔 your'e just opening the connection but never closing it

fhur, line 95:
```suggestion
            prependSql: 'SET search_path TO public',
```

fhur (approved):

fhur (approved):

PR #32, opened 2024-07-28
feat: adds support for `.count()` in the query builder

fhur:
# Why

Adds support for ".count()" and minimal support for aggregations in the query builder.


packages/backend/src/tests/e2e/nx1.test.ts
```diff
@@ -166,4 +166,10 @@ describe('n x 1', () => {
             expect(c.expected).toEqual(c.expected);
         });
     });
+
+    test('x', async () => {
```
fhur, line 170: remove

packages/queries/src/ExperimentalQueryBuilder.ts
```diff
@@ -0,0 +1,147 @@
+// ignore type errors in this file
```
fhur, line 1: we shouldn't merge this

PR #35, opened 2024-08-15
fix: Date serialization & typing

jimezesinachi:

## Why

Fixed the issue we we're having with inconsistent Date() serialization and typing

## What changed

A couple of files in `backend` & `introspect` & `react`

## Important


packages/backend/package.json
```diff
@@ -45,6 +45,7 @@
         "dotenv": "^16.3.1",
         "fast-check": "^3.19.0",
         "kysely-codegen": "^0.11.0",
+        "next": "^14.2.5",
```
fhur, line 48: Jim, we've talked about this. You can't add these dependencies so willy nilly. Just like we don't have a dependency on `express` we shouldn't have a dependency on Next.

The fact that you mark it as a `dev` dependency doesn't mean the bundler won't try to bundle it: it's part of production code now.

packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -8,6 +8,19 @@ import { composeQuery } from './composeQuery';
 import { hydrate } from './hydrate';
 import { SynthqlError } from '../../../SynthqlError';
 
+const TIME_OID = types.builtins.TIME;
+const TIMETZ_OID = types.builtins.TIMETZ;
+const DATE_OID = types.builtins.DATE;
+const TIMESTAMP_OID = types.builtins.TIMESTAMP;
+const TIMESTAMPTZ_OID = types.builtins.TIMESTAMPTZ;
```
fhur, line 15: https://discord.com/channels/1222945509266817055/1222945509740642317/1273354417638342687

fhur: No useless indirections please. Read your code before you ask for a review. 

Please make a conscious effort to build that habit.

packages/backend/src/tests/e2e/payments.test.ts
```diff
@@ -31,7 +32,7 @@ describe('e2e', () => {
             SELECT
                 payment_id,
                 amount,
-                payment_date
+                to_json(payment_date::timestamptz) as payment_date -- // TODO: fix this line
```
fhur, line 35: Please add more context. This comment is not very helpful if we come back to fix this bug in 2 weeks. We will have forgotten everything and the  lack of context just makes this useless.

fhur: It's like getting a bug report saying: "something isn't working".

packages/backend/src/tests/e2e/store-with-customers.test.ts
```diff
@@ -46,7 +44,8 @@ describe.skip('e2e', () => {
         executors: [pgExecutor],
     };
 
-    test(`${describeQuery(q)}`, async () => {
+    // TODO: fix the test by fixing the sorting function and unskip
```
fhur, line 47: now this is actually a useful TODO 👍

packages/backend/src/util/tree/fetchJsonLines.ts
```diff
@@ -0,0 +1,55 @@
+import { ReadableStream } from 'stream/web';
+
+export async function* fetchJsonLines<T = any>(
```
fhur, line 3: what are you using this for?

also why is it under "util/tree". There is nothing "tree"-ish about this.

packages/introspect/src/generate.ts
```diff
@@ -387,25 +387,34 @@ function createWellKnownDefs(): Record<string, JSONSchema> {
...
+            id: 'pg_catalog.time',
+            type: 'string',
+            format: 'time',
+            description: 'A PG time',
+        },
+        'pg_catalog.timetz': {
+            id: 'pg_catalog.timetz',
+            type: 'string',
+            format: 'time',
+            description: 'A PG timetz',
```
fhur, line 400: Update these comments to explain the format of the generated string.
Explain the caveat with the format being sometimes a little bit different. 
Explain also the caveat that PG dates have a higher precision to JS dates.

Use the venerable `['this','is','a','line'].join('\n')` format for multi string comments.

packages/react/src/fetchJsonLines.ts
```diff
@@ -6,15 +6,17 @@ export async function* fetchJsonLines<T = any>(
...
         throw new Error(
             `Failed to fetch: ${url} - Status: ${response.status} ${response.statusText}`,
         );
     }
 
     const reader = response.body?.getReader();
+
     if (!reader) {
-        throw new Error('No reader available for the response body.');
+        throw new Error('No reader available for the response body!');
```
fhur, line 19: Please avoid these cosmetic changes to files. It messes up the git history. Why does this matter?

As projects mature, finding bugs gets really tricky sometimes, and change is quite often a big predictor of bugs.

It's not that these changes are likely to introduce bugs, but if you see a file that hasn't changed in 2 years, it gives you a big indication that it's going to be safe.

fhur (changes requested):
Good progress @jimezesinachi. A few things I don't like.

 I'm leaving more feedback on discord.

packages/backend/src/util/tree/fetchJsonLines.ts
jimezesinachi: In the Next handler tests, to read the stream response. Will move to tests

fhur (approved):
LGTM 👍

PR #36, opened 2024-08-19
feat: added `createNextSynthqlHandler()` & moved the handlers to new packages

jimezesinachi:

## Why

Added `createNextSynthqlHandler()` & moved `createExpressSynthqlHandler()` & `createNextSynthqlHandler()` into their own new packages

## What changed

New packages, new functions and tests

## Important


packages/next/README.md
```diff
@@ -0,0 +1,29 @@
+# @synthql/handler-next
```
fhur, line 1: Very nice @jimezesinachi 👍 short, and easy to grok 👍

scripts/publish-minor-versions.cjs
```diff
@@ -23,6 +23,8 @@ console.log('Bumping minor version to', nextVersion);
 const packagesToUpdate = [
     '@synthql/backend',
     '@synthql/cli',
+    '@synthql/handler-express',
```
fhur, line 26: if you're going to name the packages handler-express I think you should also name the folder handler-express.

fhur (changes requested):
Very, very nice! Positively impressed by such a quick and good looking change. Only one suggestion to keep the folders consistent with the package names.

PR #37, opened 2024-08-29
fix: schema def

jimezesinachi:

## Why

## What changed

Brief summary of the changes.

## Important

reviewers.


packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -32,6 +32,8 @@ interface PgExecutorProps {
     logging?: boolean;
     /**
      * An SQL statement that will be sent before every SynthQL query.
+     *
+     * e.g `SET search_path TO "public";`
```
fhur, line 36: Is this a good example? I think messing with the search path can results in bugs. We make some assumptions about your search path being at public.

packages/cli/src/cli.ts
```diff
@@ -85,7 +89,11 @@ export function cli({
                     throw new Error(
                         [
                             `Schema definition overrides does not match the validation schema:`,
-                            ...(validateSchemaDefOverrides.errors ?? []),
+                            ...(JSON.stringify(
```
fhur, line 92: JSON.stringify is just a string. Why `?? []` and `...` ?

packages/backend/src/tests/propertyBased/tablesToSkip.ts
```diff
@@ -0,0 +1,9 @@
+export const tablesToSkip = [
```
fhur, line 1: Add a comment explaining why we skip these. We will not remember 1 year from today.

packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
jimezesinachi: Changed it to `SELECT version();`

fhur (approved):

PR #38, opened 2024-09-04
feat: added query validator `validateNestedQueriesHaveAValidRefOp()`

jimezesinachi:

## Why

Added a validator function `validateNestedQueriesHaveAValidRefOp()` that checks the shape of nested queries to make sure they have a `RefOp` where clause (i.e JOIN clause) included

## What changed

`queries` and some tests in `backend`

## Important

Also includes the small change for the CLI schema validation


packages/queries/src/QueryBuilderError.ts
```diff
@@ -0,0 +1,95 @@
...
+            'You need to add one like:',
+            '',
+            `.where({ columnName: col('rootQueryTableName.columnName') })`,
+            '',
+            'to your SynthQL query builder,',
+            'to connect (i.e. `JOIN`) the nested query to its parent query',
+            '',
+        ];
+
+        return new QueryBuilderError(type, lines.join('\n'));
```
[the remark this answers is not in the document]
fhur: Or maybe even simpler:

The table "parent" is including table "child", but "child" is missing a join predicate. Hint: are you missing `.where({some_id: "parent.some_id"})`

Where parent and child are replaces by actual table names

packages/queries/src/query.test.ts
```diff
@@ -92,24 +92,6 @@ describe('queries', () => {
         };
     });
 
-    test('Find one actor by ID', () => {
```
fhur, line 95: why did you remove tjos?

packages/queries/src/validators/validateNestedQueriesHaveAValidRefOp.ts
```diff
@@ -0,0 +1,29 @@
+import { QueryBuilderError } from '../QueryBuilderError';
+import { Query } from '../types/types';
+import { isRefOp } from '../types/isRefOp';
+
+/**
```
fhur, line 5: Lots of grammar errors here.

```suggestion
Validate that every included sub-query has at least one refOp
```

packages/queries/src/validators/validateNestedQueriesHaveAValidRefOp.ts
```diff
@@ -0,0 +1,29 @@
...
+ nested queries (includes), also has the
+ correct RefOp where clauses for each
+ */
+export function validateNestedQueriesHaveAValidRefOp<DB>(query: Query<DB>) {
+    const nestedQueries = Object.entries(query.include ?? {});
+
+    for (const [_, nestedQuery] of nestedQueries) {
+        const whereClauses = Object.entries(nestedQuery.where);
+
+        // Checking for one condition here:
```
fhur, line 16: this comment seems outdated. Please remove.

packages/queries/src/QueryBuilderError.ts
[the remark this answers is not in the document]
fhur: Concrete example:

The table movie is including table actor, but actor is missing a join predicate. Hint: are you missing `where({some_id: "movie.some_id"})` on the actor query?

packages/queries/src/query.test.ts
jimezesinachi: Two identical tests... compare it to the one just above it

fhur: Cool

packages/queries/src/QueryBuilderError.ts
```diff
@@ -0,0 +1,81 @@
...
+        // this now, is that `AnyQuery`, still lives in the backend package
+        // public query: AnyQuery,
+        public type: string,
+        public message: string,
+    ) {
+        super(message);
+        Error.captureStackTrace(this, QueryBuilderError);
+    }
+
+    static createNestedQueryMissingRefOpWhereClauseError<DB>({
```
fhur, line 16: wait why do we have this and `createNestedQueryInvalidRefOpWhereClauseError`?

packages/queries/src/validators/validateNestedQueriesHaveAValidRefOp.test.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, expect, test } from 'vitest';
+import { col } from '../col';
+import { from } from '../generated';
+
+describe('validateNestedQueriesHaveAValidRefO', () => {
```
fhur, line 5:
```suggestion
describe('validateNestedQueriesHaveAValidRefOp', () => {
```

packages/queries/src/query.test.ts
```diff
@@ -148,6 +131,7 @@ describe('queries', () => {
                 first_name: string;
                 last_name: string;
             }>;
+            filmActor: {}[];
```
fhur, line 134: this shouldn't be there? were you getting a type error?

fhur: queries with an empty select should not show up in the query result

packages/queries/src/QueryBuilderError.ts
jimezesinachi: Invalid was one I created when I was working with the idea that we could check for incorrect RefOps... might still be useful in the future, so I kept it

packages/queries/src/query.test.ts
jimezesinachi: Nope. I just saw it was missing and added it.

packages/queries/src/QueryBuilderError.ts
fhur: Don't, we can always re-build it. Unused code has a tendency to just take up space and never get used.

packages/queries/src/query.test.ts
jimezesinachi: > queries with an empty select should not show up in the query result

Ahh, I see now. That suggests we have a bug then

packages/queries/src/validators/validateNestedQueriesHaveAValidRefOp.ts
```diff
@@ -0,0 +1,23 @@
+import { QueryBuilderError } from '../QueryBuilderError';
+import { Query } from '../types/types';
+import { isRefOp } from '../types/isRefOp';
+
+/**
+ Validate that every included sub-query has at least one RefOp
+ */
+export function validateNestedQueriesHaveAValidRefOp<DB>(query: Query<DB>) {
+    const nestedQueries = Object.entries(query.include ?? {});
```
fhur, line 9: just realised: you're using `entries` but only actually using the values. Same below. Just use `Object.values`.

More attention to details please :)

fhur (approved):

PR #40, opened 2024-09-05
feat: added `name()` method to `QueryBuilder` & `name` property to `Query`

jimezesinachi:

## Why

We want users to be able to name their SynthQL queries

## What changed

Added `name()` method to `QueryBuilder` & `name` property to `Query`

## Important


packages/queries/src/query.ts
```diff
@@ -19,6 +19,7 @@ export class QueryBuilder<
     TCardinality extends 'one' | 'maybe' | 'many',
     TLazy extends true | undefined,
     TGroupBy extends string[],
+    TName extends string | undefined,
```
fhur, line 22: no need to have the name as a type parameter. This should just be a regular string.

fhur: The reason being that the name doesn't impact the output of QueryResult

packages/queries/src/query.ts
```diff
@@ -30,6 +31,7 @@ export class QueryBuilder<
         private _cardinality: TCardinality,
         private _lazy: TLazy,
         private _groupBy: TGroupBy,
+        private _name: TName,
```
fhur, line 34: i.e. this should just be a string

packages/queries/src/query.ts
```diff
@@ -439,6 +475,46 @@ export class QueryBuilder<
...
+            TGroupBy,
+            TName
+        >(
+            this._from,
+            this._where,
+            this._select,
+            this._include,
+            this._limit,
+            this._offset,
+            'many',
```
fhur, line 514: this is wrong: not all named queries return many.

fhur (approved):

PR #41, opened 2024-09-06
feat: added `hashQuery()` and used it in `useSynthql()`

jimezesinachi:

## Why

We need a way to a create consistent hashes for queries across environments

## What changed

Added the hashing function and used it in generating the endpoints in the `useSynthql()` hook

## Summary by CodeRabbit

- **New Features**
	- Introduced a `hash` property to enhance query handling and identification.
	- Added new exports for `QueryParam` and `isQueryParam` to improve query parameter management.
	- Implemented `hashQuery` utility for stable query object hashing.
	- Added `isQueryParam` function for validating query parameters.

- **Bug Fixes**
	- Updated `validateNestedQueriesHaveAValidRefOp` to exclude `hash` from the `Query` type for validation.

- **Documentation**
	- Updated base64-encoded strings for navigation and search data, potentially affecting related functionalities.


packages/react/src/useSynthql.ts
```diff
@@ -25,7 +25,15 @@ export function useSynthql<
 ): UseQueryResult<QueryResult<DB, TQuery>> {
     const { endpoint, requestInit } = useSynthqlContext();
 
-    const enrichedEndpoint = `${endpoint}/${query.from}`;
+    const hash = hashQuery(query);
```
fhur, line 28: you shouldn't calculate the hash on the react side, you should calculate it on the query builder.

packages/queries/src/hashQuery.ts
```diff
@@ -0,0 +1,55 @@
+import { AnyQuery } from './types/AnyQuery';
+
+// Copied from: https://github.com/TanStack/query/blob/353e4ad7291645f27de6585e9897b45e46c666fb/packages/query-core/src/utils.ts#L205
+/**
+ * Default query & mutation keys hash function.
+ * Hashes the value into a stable hash.
+ */
+export function hashQuery(query: AnyQuery): string {
+    return JSON.stringify(query, (_, val) =>
```
fhur, line 9: Be aware that you will need to modify the hash when you introduce query parameters.

`hashQuery(from().where({id:param('id',1)}).many())`

should give the same result as

`hashQuery(from().where({id:param('id',2)}).many())`

fhur: So probably we should do that already in this PR

packages/react/src/useSynthql.ts
```diff
@@ -25,7 +25,15 @@ export function useSynthql<
...
+
+    const enrichedEndpoint = `${endpoint}/${query.name ?? query.from}-${hash}`;
+
+    // TODO: convert a query into a new HTTP request
```
fhur, line 32: don't leave these todos hanging out there

packages/queries/src/index.ts
```diff
@@ -14,4 +13,6 @@ export * from './types/Select';
 export * from './types/Table';
 export * from './types/Where';
 export * from './types/WhereClause';
+export { col } from './col';
+export { hashQuery } from './hashQuery';
```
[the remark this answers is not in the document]
fhur: So when you build a query, you also calculate it's hash

packages/queries/src/types/QueryParam.ts
```diff
@@ -0,0 +1,5 @@
+export type QueryParam<TValue = unknown> = {
```
fhur, line 1: The name `QueryParam` is generally associated with URL query parameters. To avoid this confusion I think we should call this `QueryParameter`

packages/queries/src/types/types.ts
```diff
@@ -10,8 +11,9 @@ export type Query<DB, TTable extends Table<DB> = Table<DB>> = {
     include?: Include<DB>;
     limit?: number;
     offset?: number;
-    cardinality?: 'one' | 'maybe' | 'many';
+    cardinality?: Cardinality;
```
fhur, line 14: 👍

packages/queries/src/types/types.ts
```diff
@@ -10,8 +11,9 @@ export type Query<DB, TTable extends Table<DB> = Table<DB>> = {
...
     lazy?: true;
     groupBy?: string[];
+    hash: string;
```
fhur, line 17: This should be optional.

packages/queries/src/validators/isQueryParam.ts
```diff
@@ -0,0 +1,9 @@
+import { QueryParam } from '../types/QueryParam';
+
+export function isQueryParam(x: any): x is QueryParam {
+    return (
```
fhur, line 4: I think it's safer if we just do two checks:

1. Check that x is not null or undefined
2. Check that x.type === "synthql::param"

For example: Maybe undefined is a valid value sometimes.

packages/react/src/useSynthql.ts
```diff
@@ -25,7 +25,7 @@ export function useSynthql<
 ): UseQueryResult<QueryResult<DB, TQuery>> {
     const { endpoint, requestInit } = useSynthqlContext();
 
-    const enrichedEndpoint = `${endpoint}/${query.from}`;
+    const enrichedEndpoint = `${endpoint}/${query.name ?? query.from}-${query.hash}`;
```
fhur, line 28: Hmmm several issues here:
1. the hash might not be URL-safe
2. the hash (as it currently stands) can be _reaaaaally_ long.

packages/backend/src/query/iterateQuery.test.ts
```diff
@@ -194,13 +194,15 @@ describe('iterateQuery', () => {
                 where: {
                     film_id: 1,
                 },
+                hash: '',
```
fhur, line 197: The hash is optional, so this should not be needed. Below in the ArbitratyQueryBuilder.

packages/queries/src/QueryBuilderError.ts
```diff
@@ -16,7 +16,7 @@ export class QueryBuilderError extends Error {
         query,
         nestedQuery,
     }: {
-        query: Query<DB>;
+        query: Omit<Query<DB>, 'hash'>;
```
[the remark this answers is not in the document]
fhur: Can you remove that TODO comment?

packages/queries/src/query.ts
```diff
@@ -16,7 +18,7 @@ export class QueryBuilder<
     TInclude extends Include<DB>,
     TLimit extends number | undefined,
     TOffset extends number | undefined,
-    TCardinality extends 'one' | 'maybe' | 'many',
+    TCardinality extends Cardinality,
```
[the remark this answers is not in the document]
fhur: Yes, write unit tests, using the query builder interface. Specifically make sure that this properties hold:

1. Identical queries result in the same hash
2. Different queries result in a separate hash

Example:

```
// identical query should be equal
hashQuery(from('actor').many()) === hashQuery(from('actor').many())

// different query should not be equal
hashQuery(from('actor').many()) === hashQuery(from('films').many())
```

packages/queries/src/query.ts
```diff
@@ -67,6 +71,18 @@ export class QueryBuilder<
             cardinality: this._cardinality ?? 'many',
             lazy: this._lazy,
             groupBy: this._groupBy,
+            hash: hashQuery({
```
fhur, line 74: why not just passing `this`? is that not possible?

fhur (commented):
Good progress. Some important issues need to be solved first.

packages/queries/src/query.ts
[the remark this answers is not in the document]
fhur: @coderabbitai can you push changes to this branch?

packages/queries/src/util/hashQuery.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    // Handles Objects created by Object.create(<arbitrary prototype>)
+    if (Object.getPrototypeOf(o) !== Object.prototype) {
+        return false;
+    }
+
+    // Most likely a plain Object
+    return true;
+}
```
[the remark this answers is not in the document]
fhur: Nah Object is good enough

fhur (approved):
LGTM

fhur (approved):
LGTM

PR #43, opened 2024-09-06
feat: added query builder aliases

fhur:

## Why

People get confused about many/maybe/one

## What changed

Added all()/first()/firstOrThrow() as aliases.

### DIFF
--- a/packages/queries/src/query.ts
+++ b/packages/queries/src/query.ts
@@ -101,9 +101,40 @@ export class QueryBuilder<
     }
 
     /**
+     * Returns all rows that match the query filters.
+     *
+     * Does not override the limit set by `.limit()`.
+     *
+     * Alias for `.many()`.
+     */
+    all() {
+        return this.many();
+    }
+
+    /**
+     * Builds a query that can return 0 or 1 rows. When no rows are found,
+     * the query will return `null`.
+     *
+     * Also sets the limit to 1.
+     */
+    first() {
+        return this.maybe();
+    }
+
+    /**
+     * Builds a query that returns exactly one row. Will throw an error if no rows match
+     * the query filters.
+     *
+     * Also sets the limit to 1.
+     */
+    firstOrThrow() {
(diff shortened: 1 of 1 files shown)


packages/queries/src/query.ts
```diff
@@ -161,10 +192,7 @@ export class QueryBuilder<
     }
 
     /**
-     * Builds a query that returns exactly one row.
-     * Will throw an error if the query returns 0.
-     *
-     * Also sets the limit to 1.
+     * Alias for {@link firstOrThrow()}
```
jimezesinachi, line 195: I would think we'd want to keep this consistent with the other two, so either one of `Alias for {@link firstOrThrow()}` or `@alias {@link firstOrThrow}`

packages/queries/src/query.ts
```diff
@@ -100,10 +100,41 @@ export class QueryBuilder<
         );
     }
 
+    /**
+     * Returns all rows that match the query filters.
+     *
+     * Does not override the limit set by `.limit()`.
+     *
+     * Alias for `.many()`.
```
jimezesinachi, line 108: We could also add this particular line of comment for the other two, so "Alias for `.maybe()`" and "Alias for `one()`"

jimezesinachi (approved):
Code all good, just small things we can add to improve the documentation. Not with my machine, so it'll be easier for you to add them

packages/queries/src/query.ts
```diff
@@ -100,10 +100,41 @@ export class QueryBuilder<
         );
     }
 
+    /**
+     * Returns all rows that match the query filters.
```
jimezesinachi, line 104: We could also use the same "Builds a query that" language from the other two here, or change the other two

jimezesinachi, line 104: We could also use the same "Builds a query that" language from the other two here, or change the other two

jimezesinachi (approved):

PR #44, opened 2024-09-06
feat: moved `AnyQuery` & related types from `backend` to `queries` & updated import locations for all references

jimezesinachi:

## Why

The `AnyQuery` type is a type that is now needed to be used in other packages, so we are moving it to the common package, `queries`

## What changed

Moved `AnyQuery` and its related types, and updated the import locations for all references

## Important


fhur (approved):
LGTM, great job!

PR #51, opened 2024-09-27
feat: added array type handler & completed well known defs for all PG data types

jimezesinachi:

## Why

## What changed

Brief summary of the changes.

## Important

reviewers.

## Release Notes

- **New Features**
	- Enhanced schema generation with the addition of `excludeTablesAndViews` property, allowing users to specify tables/views to exclude.
	- Updated JSON schemas to support more complex item definitions for columns and properties.
	- New schema definition added for the `public.film` table, including the `special_features` field.

- **Bug Fixes**
	- Improved handling of PostgreSQL built-in types in the schema generation process.

- **Documentation**
	- Enhanced comments for better readability in various files.

- **Chores**
	- Added a new dependency for improved JSON schema validation.


packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
```diff
@@ -9,8 +9,6 @@ import { hydrate } from './hydrate';
 import { SynthqlError } from '../../../SynthqlError';
 
 // Use the OIDs imported from pg.types to set custom type parsers
-types.setTypeParser(types.builtins.TIME, (value) => value);
```
fhur, line 12: what happened here?

packages/introspect/src/e2e.test.ts
```diff
@@ -1,6 +1,24 @@
...
+import { generate } from './generate';
+import {
+    createTableQuery,
+    deleteTableQuery,
+    insertTableQuery,
+    selectTableQuery,
+} from './tests/sqlQueries';
+
+// Use the OIDs imported from pg.types to set custom type parsers
+types.setTypeParser(types.builtins.DATE, (value) => value);
```
fhur, line 13: Why are you putting this here?

packages/introspect/src/e2e.test.ts
```diff
@@ -81,4 +99,74 @@ describe('e2e', () => {
...
+
+    test('Add to Pagila', async () => {
+        await deleteData();
+
+        await createTable();
+
+        await insertData();
+
+        for (const [name, value] of Object.entries(await selectData())) {
+            console.log({
```
fhur, line 111: what is this for?

packages/introspect/src/e2e.test.ts
```diff
@@ -81,4 +99,74 @@ describe('e2e', () => {
...
+
+        console.log(`Record found successfully!`);
+
+        return results.rows[0];
+    } catch (err) {
+        console.error('Error finding record', err);
+    }
+};
+
+const deleteData = async () => {
```
fhur, line 160: this file is breaking the code co-location principle @jimezesinachi :

You're basically splitting functionality that is dependend on each other into two separate files. You should keep functionality that depends on each other co-located.

fhur: Honestly I would just put everything in the same file and improve the name of the file. 

We should call it something along the lines of "test serialisation of all pg types"

packages/introspect/src/e2e.test.ts
```diff
@@ -81,4 +99,74 @@ describe('e2e', () => {
                 }),
         });
     }, 100_000);
+
+    test('Add to Pagila', async () => {
+        await deleteData();
```
fhur, line 104: it's generally safer to to the cleanup in a try/finally block, something like this:

```
try {
   await createTable()
} finally {
   await dropTable();
}
```

packages/introspect/src/e2e.test.ts
```diff
@@ -81,4 +99,74 @@ describe('e2e', () => {
...
+            });
+        }
+
+        await deleteData();
+    }, 100_000);
 });
+
+const createTable = async () => {
+    try {
+        await client.connect();
```
fhur, line 124: You should re-use the same client between all these calls. No need to re-connect on every statement.

packages/backend/src/execution/executors/PgExecutor/PgExecutor.ts
jimezesinachi: Not much... I learned that TIME & TIMEZ are already being returned as string, hence no real need for specifying the custom parser

packages/backend/src/tests/generated/db.ts
```diff
@@ -57,6 +57,13 @@ export type PgCatalogNumeric = string;
  * The mpaa_rating enum from the public schema
  */
 export type PublicMpaaRatingEnum = 'G' | 'PG' | 'PG-13' | 'R' | 'NC-17';
+/**
+ * A PG text
+ */
+export type PgCatalogText1 = {
```
fhur, line 63: this is really suspicious 🤔

packages/cli/README.md
```diff
@@ -14,11 +14,13 @@ synthql generate \
     # The path where types will be generated to
     --out=src \
     # The default schema to use
+    # Option alias: `--schema`
```
fhur, line 17: Notice that most of the diff in this MR is just about this small cosmetic change. It makes it harder to review, and will make merging harder for the person who has to merge these changes back into their branch.

packages/introspect/src/generate.ts
```diff
@@ -496,78 +691,84 @@ function createWellKnownDefs(): Record<string, JSONSchema> {
...
+        },
+        'pg_catalog.xml': {
+            id: 'pg_catalog.xml',
+            type: 'string',
+            description: 'A PG xml',
         },
     };
 }
+
+function createArrayWellKnownDefs(
```
fhur, line 747: Do we really need this? This seems like it just creates an explosion of possible types.

packages/introspect/src/tests/sqlQueries.ts
```diff
@@ -0,0 +1,159 @@
+export const createTableQuery = `
```
fhur, line 1: Did you consider generating these queries dynamically, to avoid all the repetitive noise:

In essence you need:

1. A type (e.g. SERIAL PRIMARY KEY / BIGINT / BOOLEAN)
2. An example, so you can write it to the table (e. . 1 / 1 / true)

With this we can create something like:

```
const allKnownTypes:Array<Thing> = [
    {type: 'SERIAL PRIMARY KEY', examples:['1','9223372036854775807']},
    {type: 'BOOLEAN', examples: ['true,false']},
    ....
]

function getColumnName(x:Thing) {
    return `col_${normalize(x.type)}`
}
```

You can then generate the select and insert pretty easily:

```
CREATE TABLE all_data_types (
    ${allKnownTypes.map(x => ...)}.join(",")
)
```

packages/introspect/src/e2e.test.ts
jimezesinachi: To replicate the same behaviour of parsing to string that we have in the PgExecutor... that will allow for the schema matching to be correct

packages/introspect/src/e2e.test.ts
jimezesinachi: Test code.... wip

packages/backend/src/tests/generated/db.ts
jimezesinachi: Indeed. I've reworked this now, but we're still WIP

packages/backend/src/tests/generated/db.ts
```diff
@@ -60,10 +60,8 @@ export type PublicMpaaRatingEnum = 'G' | 'PG' | 'PG-13' | 'R' | 'NC-17';
 /**
  * A PG text
  */
-export type PgCatalogText1 = {
-    [k: string]: unknown;
-} & PgCatalogText2;
-export type PgCatalogText2 = string[];
+export type PgCatalogText2 = string;
+export type PgCatalogText1 = PgCatalogText2[];
```
fhur, line 64: niceeee

packages/introspect/src/generate.ts
```diff
@@ -267,6 +269,35 @@ function createTableJsonSchema(
...
+    } else if (items[items.length - 1] === '') {
+        return {
+            id: expandedType,
+            type: 'array',
+            items: createArrayDef(items.slice(0, -1).join('[]'), fullName),
+        };
+    }
+
+    // return `unknown` otherwise
+    return {};
```
fhur, line 298: 👍

fhur: I would add a log statement

fhur: using `console.warn` doens't have to be too fancy

packages/cli/src/validators/schemas.ts
```diff
@@ -67,6 +67,18 @@ export const schemaDefOverridesSchema: JSONSchemaType<
...
+                                    type: 'object',
+                                    properties: {
+                                        type: {
+                                            type: 'string',
+                                        },
+                                    },
+                                    nullable: true,
+                                    required: ['type'],
+                                    additionalProperties: true,
+                                },
```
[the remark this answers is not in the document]
fhur: @jimezesinachi ^ did you see this

packages/cli/src/validators/schemas.ts
```diff
@@ -116,6 +128,17 @@ export const schemaDefOverridesSchema: JSONSchemaType<
...
+                                                    type: 'object',
+                                                    properties: {
+                                                        type: {
+                                                            type: 'string',
+                                                        },
+                                                    },
+                                                    nullable: true,
+                                                    required: ['type'],
+                                                    additionalProperties: true,
+                                                },
```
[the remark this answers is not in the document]
fhur: also relevant here

packages/docs/static/schemas/synthql.config.json
```diff
@@ -108,6 +119,19 @@
                                                             "type": "string",
                                                             "nullable": true
                                                         },
+                                                        "items": {
```
fhur, line 122: this seems wrong 🤔 why would there be an `items` key here?

packages/introspect/src/tests/pg_types.test.ts
```diff
@@ -0,0 +1,288 @@
...
+const selectData = async () => {
+    const selectDataQuery = `
+        SELECT
+            ${allKnownTypes()
+                .map((column) => `${getColumnName(column.type)}`)
+                .join(',')}
+        FROM
+            all_data_types
+        LIMIT
+            1;
```
fhur, line 115: why limit? you should select all

packages/introspect/src/tests/pg_types.test.ts
```diff
@@ -0,0 +1,288 @@
...
+        console.log(`Table dropped successfully!`);
+
+        return results.rows[0];
+    } catch (err) {
+        console.error('Error dropping table', err);
+        throw err;
+    }
+};
+
+function allKnownTypes(): Array<Column> {
```
fhur, line 147: very nice

packages/introspect/src/tests/pg_types.test.ts
```diff
@@ -0,0 +1,288 @@
...
+                connectionString:
+                    'postgresql://postgres:postgres@localhost:5432/postgres',
+                includeSchemas: ['public'],
+                defaultSchema: 'public',
+                includeTables: [tableName],
+                outDir: '../introspect/src/tests/generated',
+            });
+
+            for (const [name, value] of Object.entries(await selectData())) {
+                const columnType = (schema.properties?.[tableName] as any)
```
fhur, line 46: This definitely needs a bit of a refactor: 

Giving a good name to this functionality makes the intention clearer.

Essentially what you want is to check that a row matches the schema, let's create one:

```
function checkRowSchema(schema:Schema, table:string, row:unknown): void {}
```

packages/docs/static/schemas/synthql.config.json
```diff
@@ -63,6 +63,17 @@
                                             "type": "string",
                                             "nullable": true
                                         },
+                                        "items": {
```
fhur, line 66: Honestly I think you should just make the whole column `type` an "any" schema e.g.

```ts
{
    schemaDefOverrides: {
        someTable: {
            type: {} <======= any
        }
    }
}
```

packages/docs/static/schemas/synthql.config.json
fhur: ok I guess its because of the reason from my previous comment

packages/introspect/package.json
```diff
@@ -32,6 +32,7 @@
     "dependencies": {
         "@apidevtools/json-schema-ref-parser": "^11.5.4",
         "@synthql/queries": "0.98.4",
+        "ajv": "^8.17.1",
```
fhur, line 35: should be a `dev` dependency, right?

packages/introspect/src/generate.ts
```diff
@@ -180,7 +188,11 @@ function createTableJsonSchema(
         // In the future we should add support for composite types.
         const type = isComposite
             ? {}
-            : { $ref: `#/$defs/${createTypeDefId(column.type)}` };
+            : isExpandedTypeAnArray(column.expandedType)
```
fhur, line 191: Let's move this logic to a function with a regular if/else. 3 cases is already enough.

packages/introspect/src/generate.ts
```diff
@@ -253,33 +265,78 @@ function createTableJsonSchema(
     };
 }
 
+function isExpandedTypeAnArray(expandedType: string): boolean {
+    return expandedType.split('[]').length > 1;
```
fhur, line 269: maybe "ends with" is safer?

packages/introspect/src/generate.ts
```diff
@@ -253,33 +265,78 @@ function createTableJsonSchema(
...
+            ? excludeTablesAndViews.length === 0
+                ? allTablesAndViews
+                : allTablesAndViews.filter(
+                      (tableOrView) =>
+                          !excludeTablesAndViews.includes(tableOrView.name),
+                  )
+            : allTablesAndViews.filter(
+                  (tableOrView) =>
+                      includeTables.includes(tableOrView.name) &&
+                      !excludeTablesAndViews.includes(tableOrView.name),
```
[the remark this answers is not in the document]
fhur: Definitely ^ those nested ? : ? : ? : are completely illegible

PR #54, opened 2024-10-08
feat: updated `shouldYield()`

jimezesinachi:

## Why

This change is the first part needed for fully supporting deferred `.lazy` queries

## What changed

The `shouldYield()` function used to determine when results are yielded from the result generator

## Summary by CodeRabbit

- **New Features**
	- Enhanced testing capabilities for various database query providers, supporting a broader range of entities.
	- New utility functions for processing nested query results, including `iterateRecursively` for traversing objects/arrays and `isDeferredResult` for type checking.
	- Introduced a function to calculate query height and a function for flattening deferred query results.
	- Added a new section in the documentation outlining SynthQL features for improved user understanding.

- **Bug Fixes**
	- Improved logic in the `shouldYield` function for better execution state evaluation.

- **Chores**
	- Updated base64-encoded strings for navigation and search data.


packages/backend/src/execution/execute.test.ts
```diff
@@ -209,7 +295,233 @@ const filmRatingProvider: FilmRatingQueryProvider = {
...
+    },
+};
+
 describe('execute', () => {
+    test('1 level deep deferred queries', async () => {
+        const store = from('store')
+            .where({
+                store_id: col('customer.store_id'),
+            })
+            .lazy()
```
fhur, line 380: as a reminder: `.lazy` has been deprecated, use `.defer()`

packages/backend/src/execution/execute.test.ts
```diff
@@ -331,6 +337,58 @@ const languageProvider: LanguageQueryProvider = {
     },
 };
 
+const paymentProvider: PaymentQueryProvider = {
```
fhur, line 340: @jimezesinachi is there a reason you added all these query providers?

fhur: I don't mind having them, I'm just wondering if I'm missing a connection between this and defer() queries

packages/backend/src/execution/execution/shouldYield.ts
```diff
@@ -8,10 +8,26 @@ function shouldYieldNode(node: ExecResultNode): boolean {
...
+        return true;
+    }
+
+    const someExecutedChildrenAreNotLazyQueries =
+        plannedChildren.length > 0 &&
+        executedChildren.some((node) => node.inputQuery.lazy !== true);
+
+    if (someExecutedChildrenAreNotLazyQueries) {
+        return true;
+    }
```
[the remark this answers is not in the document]
jimezesinachi: Very good suggestion. But in this case, we want to avoid that empty array handling, as we want empty arrays to be caught by other conditions further down

packages/backend/src/execution/execute.test.ts
```diff
@@ -289,3 +689,129 @@ describe('execute', () => {
...
+            if (x.status === 'done') {
+                setNestedValue(queryWithDeferResult, _, x.data);
+            }
+        }
+    });
+}
+
+function setNestedValue(obj: any, path: string[], value: unknown) {
+    for (let i = 0; i < path.length - 1; i++) {
+        if (!obj.hasOwnProperty(path[i])) {
```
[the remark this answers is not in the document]
jimezesinachi: Our codebase is not es2022 or later, so Object.hasOwn does not exist for us

PR #58, opened 2024-10-11
feat: fix schema sorting

jimezesinachi:

## Why
We want a stable sort of the table and column names in the generated files

## What changed
Adding some .sort() around the generate() function

## Important


fhur (approved):
LGTM

PR #60, opened 2024-10-14
feat: added `middleware()` API, types & tests

jimezesinachi:

## Why
We want to make it possible for users to pass a set of middleware that can be used to transform matching queries

## What changed
A new `middleware()` function, a `Middleware` interface, and changes to `QueryEngine.execute()`

## Important


packages/backend/src/QueryEngine.ts
```diff
@@ -30,6 +31,40 @@ export interface QueryEngineProps<DB> {
...
+     *         ...query,
+     *         where: {
+     *             ...query.where,
+     *             customer_id: context.id,
+     *         },
+     *     }),
+     * });
+     * ```
+     */
+    middlewares?: Array<Middleware<any, any>>;
```
fhur, line 67: nice!

packages/backend/src/QueryEngine.ts
```diff
@@ -96,8 +133,13 @@ export class QueryEngine<DB> {
...
     }
 
-    execute<TTable extends Table<DB>, TQuery extends Query<DB, TTable>>(
+    execute<
+        TTable extends Table<DB>,
+        TQuery extends Query<DB, TTable>,
+        TContext,
+    >(
         query: TQuery,
+        context?: TContext,
```
fhur, line 142: context should be part of `opts`

fhur: Generally speaking I try not to have too many function arguments. Why so? readability.

```ts
updateUser(123,false)
```

is way less readable than

```ts
updateUser({id:123, is_active:false})
```

If you ever get to work with a codebase that abuses positional parameters you will know what a hell it is to call 

```ts
sendEvent("user.registered",1,2,new Date(),true,true,false)
```

And how easy it is to accidentally mistype something.

packages/backend/src/QueryEngine.ts
```diff
@@ -113,7 +155,23 @@ export class QueryEngine<DB> {
             returnLastOnly?: boolean;
         },
     ): AsyncGenerator<QueryResult<DB, TQuery>> {
-        const gen = execute<DB, TQuery>(query, {
+        let transformedQuery: any = query;
```
fhur, line 158: why not make this `TQuery`?

packages/backend/src/execution/middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+    transformQuery: ({
+        query,
+        context,
+    }: {
+        query: TQuery;
+        context: TContext;
+    }) => TQuery;
+}
+
+export function middleware<TQuery = unknown, TContext = unknown>({
```
fhur, line 18: this function isn't doing anything. It's just a layer of indirection.

packages/backend/src/QueryEngine.ts
fhur: It's also easier to refactor: you can see it on this very MR: just notice all the work you had to do with refactoring the `queryEngine.execute` calls

```
return queryEngine.execute(query, {}, { returnLastOnly });
                                                            ^
                                                             |    
```

It's not necessary, just do

```ts
return queryEngine.execute(query, { returnLastOnly, context?: Context });
```

fhur:
Looks mostly what I would have expected, good job jim 👍

fhur (approved):
LGTM!

PR #61, opened 2024-10-16
feat: add ACL, `permissions` functionality

jimezesinachi:

## Why
We want to create a simple way to allow users add simple ACL controls to their queries

## What changed
A new `permissions` property in the `Query` interface, a new `.permissions()` method in the `QueryBuilder` & and a default middleware to check for the passed permissions, added to the `QueryEngine`

## Release Notes

- **New Features**
	- Introduced a new optional property `dangerouslyIgnorePermissions` to enhance permission handling in queries.
	- Added a static method `createPermissionsError` for better error reporting on permission issues.
	- New property-based tests for permission handling ensure correct behavior under various scenarios.
	- Enhanced `QueryBuilder` to support permissions throughout the query-building process.
	- Added a new `permissionsMiddleware` for validating permissions within query contexts.

- **Bug Fixes**
	- Improved robustness of middleware processing by ensuring default values for context parameters.

- **Documentation**
	- Updated documentation for various properties and methods to clarify usage and examples.

- **Tests**
	- Enhanced middleware tests and added property-based tests for more comprehensive coverage.


packages/backend/src/QueryEngine.ts
```diff
@@ -131,7 +148,34 @@ export class QueryEngine<DB> {
                 connectionString: config.url,
                 max: 10,
             });
-        this.middlewares = config.middlewares ?? [];
+        this.middlewares = [
+            ...(config.middlewares ?? []),
+            middleware<AnyQuery, AnyContext>({
```
fhur, line 153: let's move the declaration of this middleware to its own file

fhur: The implementation is also very odd to me. I need to think of a way to simplify it.

packages/backend/src/QueryEngine.ts
```diff
@@ -131,7 +148,34 @@ export class QueryEngine<DB> {
...
+                    mapRecursive(query, (node) => {
+                        if (isQueryWithPermissions(node)) {
+                            permissions.push(...(node?.permissions ?? []));
+                        }
+
+                        return node;
+                    });
+
+                    if (
+                        config.dangerouslyAllowNoPermissions ||
```
fhur, line 166: I would move this check upwards: just don't include the middleware at all if `dangerouslyAllowNoPermissions` is true.

packages/backend/src/SynthqlError.ts
```diff
@@ -135,6 +135,18 @@ export class SynthqlError extends Error {
 
         return new SynthqlError(new Error(), type, lines.join('\n'), 404);
     }
+
+    static createPermissionsError() {
```
fhur, line 139: It would be more informative if you mentioned which permission(s) are missing

packages/backend/src/QueryEngine.ts
fhur: Ok I see what's going on: the middleware is not running recursively over the queries, and so you need to it manually here.

Don't you think we should move the middleware to iterate over queries in the query tree recursively?

packages/backend/src/SynthqlError.ts
```diff
@@ -135,6 +135,18 @@ export class SynthqlError extends Error {
...
+        const type = 'PermissionsError';
+
+        const lines = [
+            'A query with a permissions list (ACL) included,',
+            'is missing matching permissions in the',
+            'context object permissions list',
+        ];
+
+        return new SynthqlError(new Error(), type, lines.join('\n'), 404);
```
fhur, line 148: 404 is not the right error code here, it's 403

fhur: > To summarize the main difference between the two, although both status codes represent access denial, 401 errors address authentication issues, and 403 errors point towards authorization problems. It is important to make this distinction as incorrectly handling these responses can leave you susceptible to exploits from malicious attackers.

Source: https://supertokens.com/blog/http-error-codes-401-vs-403#:~:text=Differences%20between%20error%20codes%20401%20and%20403,-While%20both%20HTTP&text=Authorization%3A%20A%20401%20error%20focuses,valid%20authentication%20but%20inadequate%20authorization.

packages/backend/src/execution/middleware.test.ts
```diff
@@ -54,8 +61,31 @@ describe('createExpressSynthqlHandler', async () => {
...
+        .permissions('user:read')
+        .include({
+            customer: from('customer')
+                .permissions('admin:read')
+                .where({
+                    customer_id: col('payment.customer_id'),
+                })
+                .one(),
+        })
+        .groupBy(
```
fhur, line 83: why the huge groupBy?

jimezesinachi: I'm not actually sure, but not including all of them is causing it to throw a aggregate error

packages/backend/src/execution/middleware.ts
```diff
@@ -39,3 +49,39 @@ export function middleware<TQuery = unknown, TContext = unknown>({
         transformQuery,
     };
 }
+
+export const aclMiddleware = middleware<AnyQuery, AnyContext>({
+    predicate: ({ query, context }) => {
+        const missingPermissions: string[] = [];
+
+        mapRecursive(query, (node) => {
+            if (isQueryWithPermissions(node)) {
```
fhur, line 58: I think it makes more sense to refactor this as follows:

```ts
predicate: () => true // runs on all queries
```

```ts
transformQuery: (query, context) => {
    throwIfPermissionsMissing(query, context.permissions??[])
    return query // doesn't modify the query
}
```
Notice how this makes it much easier to understand what the middleware does:

The key questions you want to ask yourself are, to what queries does this middleware apply to? all of them.

What happens when the middleware is applied to a query? it might throw an error, otherwise it won't modify the query.

packages/backend/src/execution/middleware.ts
```diff
@@ -39,3 +49,39 @@ export function middleware<TQuery = unknown, TContext = unknown>({
...
+
+            return node;
+        });
+
+        return true;
+    },
+    transformQuery: ({ query }) => query,
+});
+
+function isQueryWithPermissions(x: any): x is AnyQuery {
```
fhur, line 85: you don't need this function

packages/backend/src/SynthqlError.ts
```diff
@@ -135,6 +135,32 @@ export class SynthqlError extends Error {
...
+        contextPermissions,
+    }: {
+        node: AnyQuery;
+        missingPermissions: string[];
+        contextPermissions: string[];
+    }) {
+        const type = 'PermissionsError';
+
+        const lines = [
+            `The query '${node?.name}' with a permissions list (ACL) included:`,
```
fhur, line 151: simplify this even more:

`The query ${query.name} is missing the following permissions: ${missingPermissions}`

The first line should have most of the useful information, use the remaining lines for additional context

packages/backend/src/execution/middleware.ts
```diff
@@ -39,3 +49,39 @@ export function middleware<TQuery = unknown, TContext = unknown>({
         transformQuery,
     };
 }
+
+export const aclMiddleware = middleware<AnyQuery, AnyContext>({
```
fhur, line 53: Let's call it PermissionsMiddleware. Keep the list of names short: every name adds additional complexity.

